PR #1, opened 2017-08-23
MD5 Hash Generator

pzentner:
Wrapper for MD5 Hash Generator. Could eventually be expanded into using SHA256 and providing more cryptography support


BardurKnudsen:
Thanks for your suggestion. We also got several other suggestions for this hashing function and we are considering them for the coming releases.

BardurKnudsen:
We added this function to codeunit 1266, which now is available here on git.

JesperSchulz:
The new library functions will ship with our November update. Thank you again for your contribution!

PR #2, opened 2017-09-01
Api Mgt.

ModigeMathias:
Codeunit for Webservices integration.
Contains functionality for initializing HttpWebRequest and parsing HttpWebResponse


StanislawStempin:
Thank you for the contribution. However, there are a few problems:
1) In Extensions V2 we already have new AL types that provide HttpRequest/Response functionality so there should be no need for creating a C/AL wrapper.
2) Procedures in this contribution use .NET types as parameters. This is not allowed in this repository because such procedures would not be possible to be consumed from AL code.

Could you please have a look at the new types in AL and let us know what is the gap, if any?
https://msdn.microsoft.com/en-us/dynamics-nav/developer/devenv-restapi-overview

COD60276.TXT
```diff
@@ -0,0 +1,237 @@
...
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+
+    LOCAL PROCEDURE "--- Init Request"@6014407();
```
BardurKnudsen: What does this function do?

COD60276.TXT
```diff
@@ -0,0 +1,237 @@
...
+    BEGIN
+    END;
+
+    PROCEDURE InitWebRequest@9(Uri@6014403 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Uri";Username@6014400 : Text;Password@6014401 : Text;Domain@6014402 : Text;VAR HttpWebRequest@1000 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Net.HttpWebRequest");
```
BardurKnudsen: Either make the procedure local or remove the DotNet types from the parameter list, as the function cannot be called from the extension.

ModigeMathias:
All dotnet variables have been moved from function signatures to global variables in the last commit.

Yes, V2 has HttpWebRequest/-Response.
I created this pull request as I am unsure whether or not all .net HttpWebRequest properties can be set through V2's HttpContent and HttpHeaders classes.
Based on the description of https://msdn.microsoft.com/en-us/dynamics-nav/developer/api/httpclient-addcertificate-method it seems unclear whether or not HttpClient.AddCertificate can be used in the same way as SetTrustedCertificateValidation() from this pull request.
SetTrustedCertificateValidation() is normally used when invoking https web services with internal url and services using selfsigned ssl certficate.
If everything is already covered in V2, then this pull request is obsolete.

StanislawStempin:
As Http is one of the areas we want to support natively in AL we will implement the support that you need in case it's not already working. Could you please try the V2 for your scenario and let us know if it meets your needs?

ModigeMathias:
Ok, that sounds nice.
Then there is no need for this pull request and I believe fitgap with V2 should be posted on another forum?
Hereby closed pull request.

PR #5, opened 2017-09-04
Unix timestamp and SHA256 Generator

cvadstrup:
- Codeunit for creating unix timestamp 
- Codeunit for SHA256 cryptography support

### DIFF
--- a/SHA256.txt
+++ b/SHA256.txt
@@ -1,0 +1,60 @@
+OBJECT Codeunit 50101 SHA256
+{
+  OBJECT-PROPERTIES
+  {
+    Date=04-09-17;
+    Time=11:35:52;
+    Modified=Yes;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+            MESSAGE(CalcPaymentHMAC());
+          END;
+
+  }
+  CODE
+  {
+
+    LOCAL PROCEDURE CalcPaymentHMAC@6150652() : Text[100];
+    VAR
+      Encoding@6150616 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Text.Encoding";
+      HMACString@6150615 : Text[1024];
+      MerchantKey@6150617 : Text[50];
+      HashValue@6150614 : Text[100];
+      CharSet@6150619 : Text[30];
+    BEGIN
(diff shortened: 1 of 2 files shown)


SHA256.txt
```diff
@@ -0,0 +1,54 @@
...
+  {
+    Date=06-09-17;
+    Time=06:27:38;
+    Modified=Yes;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+            MESSAGE(CalcHMAC('SHA256 HASH Generator','iso-8859-1'));
```
BardurKnudsen: Add a test codeunit instead.

SHA256.txt
```diff
@@ -0,0 +1,54 @@
...
+  {
+
+    PROCEDURE CalcHMAC@6150652(HMACString@6150613 : Text[1024];CharSet@6150615 : Text[30]) : Text[100];
+    VAR
+      Encoding@6150616 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Text.Encoding";
+      MerchantKey@6150617 : Text[50];
+      HashValue@6150614 : Text[100];
+    BEGIN
+      Encoding := Encoding.GetEncoding(CharSet);
+      MerchantKey := '123456789';
```
BardurKnudsen: Should this be a parameter? Seems like a test value?

Unix Time.txt
```diff
@@ -0,0 +1,50 @@
...
+  {
+    OnRun=BEGIN
+            MESSAGE(FORMAT(GetUnixTime(CURRENTDATETIME)));
+          END;
+
+  }
+  CODE
+  {
+
+    PROCEDURE GetUnixTime@6150645(FromDateTime@6150614 : DateTime) UnixTimeStamp : Integer;
```
BardurKnudsen: Did you consider adding the function to codeunit 10 instead  - and write it using a similar way as the function EvaluateUnixTimestamp(Timestamp : BigInteger) : DateTime ?

BardurKnudsen:
We believe that we covered the SHA256 hashing by the functions added to codeunit 1266, and we believe that the unix time stamp can be implemented as in codeunit 10.

JesperSchulz:
The new library functions will ship with our November update. Thank you again for your contribution!

PR #6, opened 2017-09-05
Create chargelogic:xslt

jcooper01:
Function to take as input an xml file and an xsl stylesheet, apply the stylesheet to the xml file, and output a transformed document.


BardurKnudsen:
Thanks for your submission  -we are considering your request and have added a work-item for it.

JesperSchulz:
I believe we have addressed this request through #15, where we added the TryTransformXMLToOutStream function. This functionality will be made available through the November update! Thank you very much for your contribution!

PR #9, opened 2017-09-21
Wrapper for MemoryStream

hougaard:
Wrapper for MemoryStream - Can also be used for Text In/Out

### DIFF
--- a/Memory Stream.txt
+++ b/Memory Stream.txt
@@ -1,0 +1,147 @@
+OBJECT Codeunit 50777 MemoryStream
+{
+  OBJECT-PROPERTIES
+  {
+    Date=20-09-17;
+    Time=[ 7:32:02 PM];
+    Modified=Yes;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+            UnitTest();
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      MemoryStream@1160530000 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.IO.MemoryStream";
+      StreamWriter@1160530001 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.IO.StreamWriter";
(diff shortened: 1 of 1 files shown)


BardurKnudsen:
Code looks good. Is it to replace the usage of BLOB fields, e.g. TempBlob?  What are the usage scenarios?

hougaard:
This is certainly a replacement for TempBlob, but also for use with the HttpClient in cases where you're downloading binaries etc..
This is one DotNet type I use a lot.

hougaard:
And it could properly also be a workaround for the missing encoding on HttpClient :) (Because you can put encodings on in and out streams.

Memory Stream.txt
```diff
@@ -0,0 +1,147 @@
...
+  CODE
+  {
+    VAR
+      MemoryStream@1160530000 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.IO.MemoryStream";
+      StreamWriter@1160530001 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.IO.StreamWriter";
+      StreamReader@1160530002 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.IO.StreamReader";
+
+    PROCEDURE Create@1160530001(InitialSize@1160530000 : Integer);
+    BEGIN
+      MemoryStream := MemoryStream.MemoryStream(InitialSize);
```
Gallimathias: Practical would be the implementation of the empty constructor too.

Gallimathias: Besides, reading and writing would be interesting and I think dispose is very important.

JesperSchulz:
If all goes according to plan, I will take a look tomorrow and try to get the code into our BaseApp before end of week. Stay tuned!

gunnargestsson:
I Don't think this should replace TempBlob but would rather add functions like CopyToTempBlob(VAR TempBlob: Record TempBlob) and InitializeFromTempBlob(TempBlob: Record TempBlob) to this library

JesperSchulz:
Code submitted through pull request #9. Closing this pull request.

JesperSchulz:
Code will be released through November Update. Thank you for your contribution!

PR #8, opened 2017-09-06
Functions to support REST authentication

gunnargestsson:
SHA, 
JwtSecurityToken,
UTC date time,
Xml <-> Json

PR #7, opened 2017-09-06
Encryption Codeunit

Gallimathias:
I have merged the two code units of @pzentner and @cvadstrup

PR #10, opened 2017-09-26
Added hashing to Encryption management codeunit 1266

BardurKnudsen:
We have published the latest version of codeunit 1266, addressing pull requests no. #1, #5, #7 and #8 for hashing functions.

### DIFF
--- a/BaseApp/COD1266.TXT
+++ b/BaseApp/COD1266.TXT
@@ -1,0 +1,434 @@
+OBJECT Codeunit 1266 Encryption Management
+{
+  OBJECT-PROPERTIES
+  {
+    Date=;
+    Time=;
+    Version List=;
+  }
+  PROPERTIES
+  {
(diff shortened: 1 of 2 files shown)


JesperSchulz:
The functions in this pull request will will ship with our November update.

marekjz:
Hash-suite in Microsoft/AL looks very good 👍 
Some would find also very convenient to have System.Security.Cryptography.Xml.SignedXml & related classes ported to AL, to process W3C XmlDigSig (2000-2013) compliant signatures. First wish is signature verification. For only verification we could actually not necessarily load a certificate from BLOB/file/store (that only for signing), but take trusted authority thumbprint + IANA cert. chain policy OID(s) as parameters to check (eventually + some extra parameters).
Maybe also some some generic X.509 certificate and signature operations would be nice to have (e.g. certificate and signature verification).

PR #12, opened 2017-09-29
Updates to JwtSecurityTokens and added unit test

gunnargestsson:
Stop using a temporary table and move to JObject

### DIFF
--- a/COD1290.TXT
+++ b/COD1290.TXT
@@ -1,0 +1,337 @@
+OBJECT Codeunit 1290 SOAP Web Service Request Mgt.
+{
+  OBJECT-PROPERTIES
+  {
+    Date=;
+    Time=;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      BodyPathTxt@1001 : TextConst '@@@={Locked};ENU=/soap:Envelope/soap:Body';
+      ContentTypeTxt@1000 : TextConst '@@@={Locked};ENU="multipart/form-data; charset=utf-8"';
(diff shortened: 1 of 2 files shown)


JesperSchulz:
I can't seem to see the differences. Something is not quite right in this pull request. Could you try again? It's hard to see the differences you want to make.

gunnargestsson:
The last four functions all need to be [External].  I have also removed "Standard Text" as a temporary table (it will not work in all cases) and use JObject instead as the temporary storage for the token claims.

JesperSchulz:
Your suggestions have functions, which have .NET parameters. Those, as you know, cannot be used from V2 extensions. I could not see any other improvements over the code we have in repo. I'm hence closing this pull request without making any changes. Feel free to correct your code and reopen this pull request for further processing.

gunnargestsson:
There are exceptions and this in one of them. This dotnet object has been mapped to AL type and in json management this dotnet type is already external and accessible by Extension V2

Kveðja/Kind regards,
Gunnar Þór Gestsson

JesperSchulz:
I see! Case re-opened :-) We will have to re-triage it! Thanks for clarifying!

StanislawStempin:
Keep in mind that new AL types are not really "mapped" to .NET types. You should treat them as completely new AL types. It will not be possible to call the methods from C/AL Open Library that have DotNet types as part of the signature even though similar types might exist in AL.

JesperSchulz:
Based on Stan's comment, I will close the pull request again, as the suggested code changes won't work for V2 extensions. Feel free to refactor and we'll reconsider! Thanks for all your contributions! They are very much appreciated!

gunnargestsson:
I will refactor and submitt another pull request.  But I suggest you take a look at Codeunit 5459 that has external function using DotNet variable.

JesperSchulz:
You're right! We have a lot of external functions in that codeunit with .NET parameters! That is a mistake! I will investigate how those got there. We probably have to make those functions internal or refactor. Thanks for pointing that out!

PR #17, opened 2017-10-13
Using Json text variable instead of using temporary table

gunnargestsson:


JesperSchulz:
This pull request was addressed through pull request #24.

PR #19, opened 2017-10-18
Suggest a new formatting function for COD10

gunnargestsson:


JesperSchulz:
I will take a look at this one! Looks doable!

JesperSchulz:
For some reason the tests fail on my machine. Didn't have time to look into details, but for whatever reason, the TypeHelper.Format(Date,'yyyy-MM-dd',''); returns the previous day, instead of the current day. Will have to look into that. Other than that, everything looks good!

gunnargestsson:
On NAV 2017, 10.0.14199 (on a remote computer) this does not compile, I am getting "The function call was ambiguous. No matching method was found.".  On my local computer on NAV 2017, 10.0.18197 this compiles.  Any idea why this fails ?  Operating system or NAV?

gunnargestsson:
I had them do a Windows Update and now everything compiles.  Running on 2012R2 server.

gunnargestsson:
Can you look at this thread ?

https://www.yammer.com/dynamicsnavdev/#/Threads/show?threadId=970623713

JesperSchulz:
We are looking at the discussion you've mentioned. We will wait for a conclusion before going forward with this pull request.

gunnargestsson:
I believe that discussion proves that the String.Format function is working perfectly so it should be safe to add it.

JesperSchulz:
I'm not sure I agree. You are now aware of the difference between a .NET DateTime and DateTime, but a developer who is not, will "loose" a couple of hours depending on which timezone he/she is in when using this format function - that would be a serious bug farm! Until that is resolved, I will part this request. Shouldn't be for long...

JesperSchulz:
The bug was rejected. The .NET format function is for formatting .NET types, not NAV types. Using this formatting method on NAV types, has some unexpected behavior and we will hence not add this function to the library. It would be har to understand, why formatting a DateTime results in a change of time.

PR #20, opened 2017-10-20
Added an intermediate page for accessing camera in V2 extensions

defr0std:
A request is made to use camera feature in V2 extensions.
https://github.com/Microsoft/AL/issues/215

The problem is that Camera API requires .NET variable which can be used only on pages. Therefore the proposed solution is to add an intermediate page which would access the camera and provide the taken picture back to the caller as a BLOB.

This solution requires the same number of clicks for the picture to be taken, but shows an intermediate form in the process. On one hand, this form could be a bit distracting, but on the other hand, this form will also report an error message if the camera is not available, so there is no need to check for that in every use case of the camera API and hide the buttons.

The form can also be customized by the extension to provide more instructions to the user.

This is how a picture could be taken using V2 code:
```JavaScript
pageextension 50108 CustomerListExt extends "Customer Card"
{
    actions
    {
        addafter(Approval)
        {
            action(ChangePicture)
            {
                ApplicationArea = All;

                trigger OnAction();
                var CameraPage : Page "Camera Interaction";
                    PictureStream : InStream;
                begin
                    CameraPage.AllowEdit(true);
                    CameraPage.Quality(100);
                    CameraPage.EncodingType('PNG');

                    CameraPage.RunModal();

                    if (CameraPage.GetPicture(PictureStream)) then begin
                        Image.ImportStream(PictureStream, CameraPage.GetPictureName());
                    end;
                end;
            }
        }
    }
}
```

P.S. This is a bit non-standard request - no tests are added since it's all integration API, no codeunits either :)


JesperSchulz:
This is a fine little functionality, but it needs some refactoring. We should not have this much code on a page. We will take a closer look and do a little refactoring and re-design.

dennisfrostlander:
@JesperSchulz Thanks. Also got some UX feedback that it would be nice to have only one OK button on the intermediate page instead of OK/Cancel. Could you please also have a look that while refactoring?

BaseApp/PAG99999.txt
```diff
@@ -0,0 +1,108 @@
...
+    SourceTableTemporary=Yes;
+    OnOpenPage=BEGIN
+                 CameraAvailable := CameraProvider.IsAvailable();
+
+                 IF CameraAvailable THEN BEGIN
+                   CameraOptions := CameraOptions.CameraOptions;
+                   CameraOptions.Quality := Quality;
+                   CameraOptions.AllowEdit := AllowEdit;
+
+                   CASE EncodingType OF
```
MarcHansenMicrosoft: Try using Evaluate to get the string value of the enumeration.

BaseApp/PAG99999.txt
```diff
@@ -0,0 +1,108 @@
...
+  {
+    Editable=No;
+    CaptionML=ENU=Camera;
+    SourceTable=Table99998;
+    PageType=StandardDialog;
+    SourceTableTemporary=Yes;
+    OnOpenPage=BEGIN
+                 CameraAvailable := CameraProvider.IsAvailable();
+
+                 IF CameraAvailable THEN BEGIN
```
MarcHansenMicrosoft: Better do 
IF NOT CameraAvailable THEN 
  EXIT;

BaseApp/PAG99999.txt
```diff
@@ -0,0 +1,108 @@
...
+      CameraOptions@1000 : DotNet "'Microsoft.Dynamics.Nav.ClientExtensions, Version=11.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.Microsoft.Dynamics.Nav.Client.Capabilities.CameraOptions";
+      CameraAvailable@1002 : Boolean INDATASET;
+
+    EVENT CameraProvider@1001::PictureAvailable@10(PictureName@1001 : Text;PictureFilePath@1000 : Text);
+    VAR
+      File@1002 : File;
+      InStream@1003 : InStream;
+      OutStream@1004 : OutStream;
+    BEGIN
+      File.OPEN(PictureFilePath);
```
MarcHansenMicrosoft: Use blob.Import(filename)

BaseApp/TAB99998.txt
```diff
@@ -0,0 +1,37 @@
...
+  {
+  }
+  FIELDS
+  {
+    { 1   ;   ;Quality             ;Integer        }
+    { 2   ;   ;AllowEdit           ;Boolean        }
+    { 3   ;   ;EncodingType        ;Option        ;OptionString=JPEG,PNG }
+    { 4   ;   ;MediaType           ;Option        ;OptionString=Picture,Video,AllMedia }
+    { 5   ;   ;SourceType          ;Option        ;OptionString=Camera,PhotoLibrary }
+    { 6   ;   ;TakenPicture        ;BLOB           }
```
MarcHansenMicrosoft: Just Call them Picture and PictureName

BaseApp/TAB99998.txt
```diff
@@ -0,0 +1,37 @@
...
+    { 7   ;   ;TakenPictureName    ;Text250        }
+  }
+  KEYS
+  {
+    {    ;Quality                                 ;Clustered=Yes }
```
MarcHansenMicrosoft: If you end up saving this you should have a better key. A GUID will do.

BaseApp/PAG99999.txt
```diff
@@ -0,0 +1,108 @@
...
+                       CameraOptions.SourceType := 'PhotoLibrary';
+                     SourceType::Camera:
+                       CameraOptions.SourceType := 'Camera';
+                   END;
+
+                   CameraProvider := CameraProvider.Create();
+                   CameraProvider.RequestPictureAsync(CameraOptions);
+                 END;
+               END;
+
```
MarcHansenMicrosoft: Us QueryClose to return the LookUpOK if you got a picture.

MarcHansenMicrosoft:
Please make sure to update the usage sample to reflect the changed code.

BaseApp/TAB99998.txt
```diff
@@ -0,0 +1,37 @@
...
+  }
+  FIELDGROUPS
+  {
+  }
+  CODE
```
MarcHansenMicrosoft: A small helper function SavePictureToStream would be good.

BaseApp/PAG99999.txt
dennisfrostlander: Moved properties to the page itself, removed options.

BaseApp/PAG99999.txt
dennisfrostlander: Done.

BaseApp/PAG99999.txt
dennisfrostlander: Done

BaseApp/TAB99998.txt
dennisfrostlander: Not relevant anymore.

BaseApp/TAB99998.txt
dennisfrostlander: Table removed

BaseApp/PAG99999.txt
dennisfrostlander: Not possible, as discussed. Instead added GetPicture(stream: InStream): boolean, as agreed.

JesperSchulz:
Addressed through pull request #44. Case closed.

dkatson:
@JesperSchulz i've just tested this in wave 2 with iPhone and found that 
`if (CameraPage.GetPicture(PictureStream))`
shows the import screen to take the picture once again.

So `CameraPage.RunModal()`
gives me this screen
image

and then I see 
image

That's because `CameraPage.GetPicture` have `FileManagement.BlobImport()` inside
image

And `FileManagement.BlobImport()` has `UploadIntoStream`
image

which opens this new step in the iPhone.
I found this interesting "remark" in the documentation https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/file/file-uploadintostream-method

image

But i think it's not longer true...

So, you need to change "Camera iteraction".GetPicture business logic.

JesperSchulz:
@dkatson, thanks for letting me know! I'll see what i can do ;-)

srkkoneti:
@JesperSchulz, We have tested the camera feature in Wave2 the same issue occurring, After uploading screen, showing blank import screen and if we cancel the import screen, we are getting following error
Media

JesperSchulz:
We've actually not been able to get around to this issue yet. There are currently quite a few more pressing tasks we need to attend to, I'm afraid. But I've created a bug for this and we should be able to get to it soon. If you'd like to suggest a fix, please go right ahead :-) That might just speed up things! Otherwise please have a little more patience.

dkatson:
Well, that was my workaround. 

    local procedure TakePicture(var TempPurchLine: Record "Purchase Line" temporary);
    var
        //CameraInteraction: Page "Camera Interaction";
        PictureStream: InStream;
        FileManagement: Codeunit "File Management";
        TempBlob: Codeunit "Temp Blob";
        FileName: Text;
    begin
        //CameraInteraction.AllowEdit(true);
        //CameraInteraction.Quality(100);
        //CameraInteraction.EncodingType('PNG');
        //CameraInteraction.RunModal();
        //if (CameraInteraction.GetPicture(PictureStream)) then begin
        //    TempPurchLine."AIR Item Photo".ImportStream(PictureStream, CameraInteraction.GetPictureName());
        //end;

        FileManagement.BLOBImport(TempBlob, FileName);
        TempBlob.CreateInStream(PictureStream);
        TempPurchLine."Item Photo".ImportStream(PictureStream, FileName);
    end;

dkatson:
So, just completely ignore Page "Camera Interaction". It works, but not so sexy as with "Camera Interaction" :)

JesperSchulz:
Module extracted. Please take it for a spin :-) https://github.com/microsoft/ALAppExtensions/tree/master/Modules/System/Camera%20and%20Media%20Interaction

srkkoneti:
@JesperSchulz , Could you please let us know in which build it was included

JesperSchulz:
The code went in just before Christmas on the 22nd of December. If you take the latest preview build of version 16, you should get it. The code was also backported to 15, but release usually takes a little while. Expect the change to be released by EOM.

PR #21, opened 2017-10-25
Small but usefull functions

gunnargestsson:


JesperSchulz:
These functions are indeed small :-) The leapyear one fits perfectly into COD10 and will be added there. The two numeric ones do not really need wrappers and are one-liners in the first place. I'm not sure we need library functions for those two.
Could you explain the EndOfLineString function? Doesn't it just return the string \r\n? Our product only runs on windows and hence this is the only thing this function returns, or am I wrong? What is the use case for this one? I'd be happy to add it to the library, once I understand what we need it for :-)

gunnargestsson:
Well, the wrappers for DELCHR are for one line, true, but putting a name to that complex DELCHR combination is the goal.

If I am writing to a Stream I usually need to add EndOfLineString and this is the best way.

Stream.WRITETEXT(Line + TypeHelper.NewLine)

other options would be to add a text variable and do

VAR CrLf: Text[2]
CrLf[1] := 13;
CrLf[2] := 10;
Stream.WRITETEXT(Line + CrLf)

JesperSchulz:
Addressed through pull request #23. Closing this pull request.

PR #25, opened 2017-11-06
Stream support for TempBlob

gunnargestsson:
I like to use TempBlob to stream text in and out.
Already built in functions to stream text out, adding function to stream text in.
This change makes TempBlob easy to use when working with text.  Example...

    TempBlob.StartReadingTextLines(TEXTENCODING::Windows);
    WHILE TempBlob.MoreTextLines DO BEGIN
      TextLine := TempBlob.ReadTextLine;
      IF STRLEN(TextLine) >= 3 THEN
        CASE COPYSTR(TextLine,1,3) OF
          'UNA','UNB':
            Header += TextLine + NewLine;
          'UNZ':
            Footer := TextLine + NewLine;
          'UNH':
            BEGIN
              IF ResultTempBlob."Primary Key" > 0 THEN
                ResultTempBlob.INSERT;
              ResultTempBlob."Primary Key" += 1;
              ResultTempBlob.StartWritingTextLines(TEXTENCODING::Windows);
              ResultTempBlob.WriteTextLine(TextLine + NewLine);
            END;
          ELSE
            ResultTempBlob.WriteTextLine(TextLine + NewLine);
        END;
    END;
    ResultTempBlob.INSERT;

    WITH ResultTempBlob DO BEGIN  
      SETAUTOCALCFIELDS(Blob);
      IF FINDSET THEN REPEAT
        TempBlob.StartWritingTextLines(TEXTENCODING::Windows);
        TempBlob.WriteTextLine('DELIMITERS(039,043,058)' + NewLine);
        TempBlob.WriteTextLine(Header);
        TempBlob.WriteTextLine(ResultTempBlob.ReadAsText(NewLine,TEXTENCODING::Windows));
        TempBlob.WriteTextLine(Footer);
        ServerTempFileName := 
          FileMgt.CombinePath(
            FileMgt.GetDirectoryName(NameValueBuffer.Name),
            FileMgt.GetFileName(FileMgt.ServerTempFileName(DELCHR(FileExtension,'<','.'))));
        FileMgt.BLOBExportToServerFile(TempBlob,ServerTempFileName);
      UNTIL NEXT = 0;
      DELETEALL;
    END;

### DIFF
--- a/BaseApp/TAB99008535.TXT
+++ b/BaseApp/TAB99008535.TXT
@@ -1,0 +1,207 @@
+OBJECT Table 99008535 TempBlob
+{
+  OBJECT-PROPERTIES
+  {
+    Date=06.11.17;
+    Time=13:35:23;
+    Modified=Yes;
+    Version List=NAVW111.00;
+  }
+  PROPERTIES
+  {
+    CaptionML=[ENU=TempBlob;
+               ISL=TempBlob];
+  }
+  FIELDS
+  {
+    { 1   ;   ;Primary Key         ;Integer       ;CaptionML=[ENU=Primary Key;
+                                                              ISL=AÐallykill] }
+    { 2   ;   ;Blob                ;BLOB          ;CaptionML=[ENU=Blob;
+                                                              ISL=BLOB] }
+  }
+  KEYS
+  {
+    {    ;Primary Key                             ;Clustered=Yes }
+  }
+  FIELDGROUPS
+  {
+  }
+  CODE
+  {
+    VAR
+      GlobalInStream@1000 : InStream;
+      GlobalOutStream@1012 : OutStream;
+      ReadLinesInitialized@1001 : Boolean;
+      NoContentErr@1002 : TextConst 'ENU=The BLOB field is empty.;ISL=The BLOB field is empty.';
+      UnknownImageTypeErr@1003 : TextConst 'ENU=Unknown image type.;ISL=Unknown image type.';
+      XmlCannotBeLoadedErr@1004 : TextConst 'ENU=The XML cannot be loaded.;ISL=The XML cannot be loaded.';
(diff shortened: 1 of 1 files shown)


BaseApp/TAB99008535.TXT
```diff
@@ -0,0 +1,207 @@
...
+    VAR
+      ContentLine@1003 : Text;
+    BEGIN
+      IF NOT MoreTextLines THEN
+        EXIT('');
+      GlobalInStream.READTEXT(ContentLine);
+      EXIT(ContentLine);
+    END;
+
+    PROCEDURE WriteTextLine@14(Content@10034985 : Text);
```
JesperSchulz: Wrong ID range

BaseApp/TAB99008535.TXT
```diff
@@ -0,0 +1,207 @@
...
+    BEGIN
+      IF NOT WriteLinesInitialized THEN
+        ERROR(WritingTextLinesNotInitializedErr);
```
JesperSchulz: I prefer to initialize directly. We do that for read as well. Will default to same encoding as when writing. We should not use errors, which are for developers, if we can avoid it. Errors will surface to the user, who will not know what to do about it.

BaseApp/TAB99008535.TXT
```diff
@@ -0,0 +1,207 @@
...
+    Version List=NAVW111.00;
+  }
+  PROPERTIES
+  {
+    CaptionML=[ENU=TempBlob;
+               ISL=TempBlob];
+  }
+  FIELDS
+  {
+    { 1   ;   ;Primary Key         ;Integer       ;CaptionML=[ENU=Primary Key;
```
JesperSchulz: Don't submit ML captions.

JesperSchulz:
Pull request addressed through #31.

PR #26, opened 2017-11-06
New function for Codeunit 1297

gunnargestsson:
I know that this is not directly adding a dotnet code but we need this function to be able to speak to a NAV OData service from C/AL.
Adding function: AddBasicAuthentication
Pretty please...


JesperSchulz:
This pull request was addressed by #32.

PR #27, opened 2017-11-08
Mediator will allow a "Message Bus" in NAV

gunnargestsson:
One App might want to send a standard message or receiva a standard message from another App.  Using this method will allow us to have a "Message Bus" in NAV/TENERIFE and Apps can communicate without having to know about each other code.

### DIFF
--- a/Mediator.txt
+++ b/Mediator.txt
@@ -1,0 +1,79 @@
+OBJECT Table 705 Mediator Argument
+{
+  OBJECT-PROPERTIES
+  {
+    Date=;
+    Time=;
+    Version List=;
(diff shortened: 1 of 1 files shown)


grobyns:
Hi Gunnar,

thank you for the suggestion however we think this is too generic a solution to include in the base code and will reject this pull request. It will be very hard for subscribers to know what they are subscribing to and the code will become very hard to read. In addition, as all subscribers will be called every time, each subscriber will have to determine if it should act or not. If any subscriber fails to do so the consequences could be very bad. 

Best regards,
Gert

PR #30, opened 2017-11-16
New Path Helper

DanielGoehler:
The Function from **System.IO.Path** in that new Codeunit are not accessing the local filesystem, but there are very useful to e.g. to separate file name from file extension.

### DIFF
--- a/BaseApp/COD50001.TXT
+++ b/BaseApp/COD50001.TXT
@@ -2,8 +2,8 @@ OBJECT Codeunit 50001 Path Helper
 {
   OBJECT-PROPERTIES
   {
-    Date=16.11.17;
-    Time=00:00:00;
+    Date=;
+    Time=;
     Version List=;
   }
   PROPERTIES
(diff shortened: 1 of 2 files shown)


BaseApp/COD50001.TXT
```diff
@@ -0,0 +1,49 @@
+OBJECT Codeunit 50001 Path Helper
```
forki: is it normal to add CDUs to 50xxx? isn't that what most people use to add customer stuff?

DanielGoehler: From https://github.com/Microsoft/cal-open-library/blob/master/README.md
> 3. Object IDs should be created within the range 50K-99K.

JesperSchulz: We're going to change them when processing the request.

JesperSchulz: Code accepted. I will add the requested functions to COD419. Thanks for your contribution!

JesperSchulz: Requested functions were added to COD419. Actually, most of the requested functions already existed, but weren't exposted. They are now. For all changes, please refer to #39.

JesperSchulz:
Thank you for your contributions. Case closed :)

PR #33, opened 2017-11-19
Need information from the system permission into my AL app.  

gunnargestsson:
Permission Table is not visible to the AL app and I need to ask if a user has access to a specific table by default in the system.
New code is the function 

PROCEDURE GetTablePermissionFromApplication@29(TableId@1000 : Integer;VAR TablePermissionBuffer@1001 : Record 9800);

### DIFF
--- a/BaseApp/COD9002.TXT
+++ b/BaseApp/COD9002.TXT
@@ -1,0 +1,425 @@
+OBJECT Codeunit 9002 Permission Manager
+{
+  OBJECT-PROPERTIES
+  {
+    Date=;
+    Time=;
+    Modified=Yes;
+    Version List=;
+  }
+  PROPERTIES
+  {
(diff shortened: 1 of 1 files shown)


PredragMaricic:
Hey Gunnar,
Why can't we use READPERMISSION or WRITEPERMISSION in this case? What exactly is advantage if we use your function?
Thanks for your contribution,
Predrag

gunnargestsson:
I could do that in most cases but this is more detailed.  For example the WRITEPERMISSION needs both Insert and Modify to return true - if I remember correctly.

gunnargestsson:
We now have direct access to Permission table from AL.  Not needed any more.

PredragMaricic:
Agree. Thanks for suggestion.

PR #34, opened 2017-11-19
New system format for type helper

gunnargestsson:
Could not verify the tests, I don't have the latest tests and objects used in this test codeunit are not available to me.


grobyns:
Hi Gunnar,

The Format and FormatUtcDatTime function you are adding here have some issues:
- we already have FormatDate, FormatDateTime, GetFormattedCurrentDateTimeInUserTimeZone. We really don't want another function that does almost the same thing. Is it possible to create 1 date formatting function that does all we need? We would have to keep the current function signatures so as not to break anyone but those should then be one-liners with all logic placed in a single function.
- When the variant passed in is a date only you add the current time to make a DateTime. This is odd behaviour. why would you need the current time as opposed to 00:00:00?
- The ELSE part will try and format anything passed in. How would this code react to someone passing in a record for example? I think it would be better to have a date/datetime specific format function and if we need a more generic format function where the built in al FORMAT is not sufficient, that one should only be allowed to pass types known to DotNet.
- The tests are not sufficient. All they seem to prove is that the built in al FORMAT function is sufficient and we really don't need to use DotNet to format dates at all.

Cheers,
Gert

gunnargestsson:
Hi Gert,

There is a difference between FORMAT(NavDateTime,0,2) and FORMAT(NavDateTime,0,9).  The first will return datetime in the user timezone where the later will return datetime in UTC timezone.  We currently don't have any function in TypeHelper to return the NavDateTime in UTC timezone.

If I use 00:00:00 then we will not properly convert the Date to local time zone in the FormatDateTime function.  By using TIME the result will be correct.

If soneone passes a variable that does not map to known DotNet variable type an error vill occur.  Same will happen if you try to use FORMAT(Customer) in C/AL where Customer is a record.  We can't stop people from burning by taking away the fire.

In the Data Exchange Framework we are using DotNet formatting and that is why I am asking for these functions.  If you where using C/AL formatting in the Data Exchange Framwork then we would not need these functions.  The option is up to you.  Rewrite the existing code or add formatting support to existing code.  The original TypeHelper DotNet formatting functions where created because of this.  I do thing we need to go all the way since we started to go down this road.

grobyns:
Hi Gunnar,

re the format difference:
I’m not saying the function isn’t needed, I’m saying we want the implementation to be cleaner. The logic should be put in one place. 

re the 00:00:00 time issue: 
good point, but your approach is also not correct imo. When I pass in a date (no time) and format it I expect that date to be returned. Depending on when the function is called the date returned could be different. We need a solution where the date returned does not depend on time at all. If time is important then it should be provided (i.e. pass in a datetime in the first place).

re the formatting of unknown types: This is not correct. FORMAT(Customer) will not throw an error. 

re data exchange framework: I don't know the in's and out's of the data exchange framework but I could not find anywhere where we use the typehelper format functions to format dates. Could you let me know where we use it so I can take a look? If you have an easy repro that would also help.

I'll repeat my ask/suggestion from before. Move all logic to a single date formatting function, ensure that function can do all the necessary formatting required for dates and turn all the other functions to one-liners. We'll have a cleaner and easier implementation to use and maintain with no duplication.

Cheers,
Gert

gunnargestsson:
Hi

So back to the "why" I need this functionality.  Look at table 1223, fields 5,6 and 7.  When I am exporting via data exchange I need to format the date according to the values in these fields.  That is why I am asking for this dotnet formatting function.

Any other solution to solve this task will be fine.

grobyns:
Hi Gunnar,

As I said, I’m no expert on the data exchange framework but AFAIK these options are there so you can specify them when importing data. If your external data is in a specific format then you can use these fields to have NAV evaluate it correctly. I don’t think it was meant to format outgoing data but I guess it could be used that way. If I remember correctly (Big IF) there should be an option to provide a nav format string to ensure the data is formatted correctly on export. Could you use that instead?

br/Gert

gunnargestsson:
I don't think there is a NAV format string in there and I believe that we should support this since this is in there.

grobyns:
Hi Gunnar,

What I am saying is that you can add normal nav format strings in the “Data Format” field (leave the culture field blank).
To get the XML format (9) you could add ‘<Standard Format,9>’ for example. Or you can add something like ‘<Day,2>-<Month,2>-<Year> <Hours24>.<Minutes,2>’.

AFAIK the code will use the FORMAT function and the format string when exporting your data.

br/Gert

gunnargestsson:
Sure that is a possibility.  What will the user experience be when you expect dotnet format while importing and nav format while exporting?

Kind Regards,
Gunnar Gestsson

grobyns:
Hi Gunnar,

AFAIK a data exchange definition is either import or export, not both but in any case you can always duplicate the data definition and have it use separate column definitions with different format/evaluate rules.

Since it seems the scenario is already supported I will close this request.

Cheers,
Gert

gunnargestsson:
Hi Gert

By all means challenge me but I don't like that you close the request without my consent.  You offered nothing in you comments that I did not already know about and I have the business case to support my request.

doctorpartlow:
Will Gert retort with a dramatic response or will Gunnar reign victorious and provide a valuable tool for the coming releases of Microsoft Dynamics business or NAV or whatever they're calling it today? Find out in the next comment of New system format for type helper

gunnargestsson:
The doctor is always right!

Kind regards,
Gunnar Gestsson

PR #35, opened 2017-11-20
Json Writer and Reader wrappers

gunnargestsson:


StanislawStempin:
Hi Gunnar, 
Could you provide some background on why you need these functions? Why not use the new Json* types directly?
Thanks,
Stan

gunnargestsson:
Hi Stan
In my workshop my students had a task to create and read json.  I simply asked them.  They felt this method of creating json would be the easiest one and wanted me to put the code here.

StanislawStempin:
Hi Gunnar, This library is meant for unblocking things not possible to achieve in Extension V2 due to lack of .NET interop. We would need to see specific scenarios that are not possible to achieve in V2 using the new Json types.

gunnargestsson:
Hi Stan
Have any of the issues with the AL types been fixed or do you still need to apply strange (dotnet) methods to get the expected results?

StanislawStempin:
Which specific issues do you refer to? You can see the current status of all issues on the AL GitHub. Once they are fixed in our builds they get marked with 'ships-in-next-update'. Otherwise it means we didn't get to fixing them yet.

gunnargestsson:
Well, this one is marked by design (copied from dotnet).  Still - in my view - unwanted behaviour for AL.
https://github.com/Microsoft/AL/issues/900

StanislawStempin:
We will be addressing #900. Apart from that one scenario (which BTW works as expected when using the Clear() AL method) are there any other blockers in the new native types that require this new library?

gunnargestsson:
This new library is just a more natural (human) way of creating json.  From the top to the bottom.  It is not as easy for the human mind to think in objects where you create an object and then put in somewhere in the file.  That is why I put this in here.

strophanthus:
Hello gunnar, this is quite helpful. I was wondering if the text writer initialize could be modified somewhat so that the indentation formatting could be optional?

JesperSchulz:
Hi Gunnar!

I wish you a happy new year :-)

Sorry about the long wait - we've been quite busy before the Holidays and I hence had to park this request for a while. But now the wait is over and I'll pick it up before end of week! It's my belated Christmas present for you ;-)

JesperSchulz:
Code added to repo through commit d96417d5dae3fcd361b5cef31202a17fb2b3dcfd. The code was slightly modified, as the read functionality already was implemented on the Json Buffer table. Also, I've added a function to skip formatting. I hope the provided solution does the trick! If not, please reopen. Thank you very much for your contribution.

PR #37, opened 2017-11-28
Need to support base64 in the hash functions

gunnargestsson:
Was testing this with Azure REST API

### DIFF
--- a/BaseApp/COD1266.TXT
+++ b/BaseApp/COD1266.TXT
@@ -367,7 +367,7 @@ OBJECT Codeunit 1266 Encryption Management
     END;
 
     [External]
-    PROCEDURE GenerateHash@39(InputString@1000 : Text;HashAlgorithmType@1001 : 'MD5,SHA1,SHA256,SHA384,SHA512') : Text;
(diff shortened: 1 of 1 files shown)


JesperSchulz:
I will look into this one in the upcoming days. The request seems valid, but I might change the way it is implemented slightly, as I'd like to avoid adding a new parameter to existing functions.

gunnargestsson:
Please do.  I need this functionality to be able to use Azure REST API

JesperSchulz:
Would it be possible to add a few tests? I need some code to confirm I get the behavior right of the functionality you are requesting.

gunnargestsson:
Possible, sure.  Let's try to find time...

JesperSchulz:
;-) I know the feeling. Unfortunately I cannot check in new code without test coverage and I'm a little short on time myself. I hope you do find the time, then I'll prioritize this request and hopefully make it in time for the February release.

JesperSchulz:
Almost there. With test and all. But I'm not very fond of having to specify the key encoding in the function. Could you not convert the encoding of the key from Base64 to UTF8 before calling? Or does that not fly?

JesperSchulz:
I will upload my suggestion, then you can improve on that if it does not fulfill all your needs :-)

JesperSchulz:
I've addressed most requests through pull request #50. Please verify that the changes cover your needs. If not, please make a new request and add the code you need (incl. tests), and I'll look into it once more.
I wish you a merry Christmas and a very happy new year!

PR #38, opened 2017-11-28
Content function for TempBlob

gunnargestsson:
The functions in Codeunit 1266 require a text import.  I also need to be able to get length and hash for binary content

### DIFF
--- a/BaseApp/TAB99008535.txt
+++ b/BaseApp/TAB99008535.txt
@@ -178,6 +178,59 @@ OBJECT Table 99008535 TempBlob
       EXIT(Xml);
     END;
 
+    [External]
+    PROCEDURE GetContentLength@20() : Integer;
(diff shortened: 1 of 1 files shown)


BardurKnudsen:
We have implemented the following:
Blob fields will get a LENGTH method, so you would just call e.g. Customer.Picture.LENGTH;
Cod1266 gets a new function: 
  PROCEDURE GenerateHashFromStream@33(InStr@1000 : InStream;HashAlgorithmType@1002 : 'MD5,SHA1,SHA256,SHA384,SHA512') : Text;
Then you just create an instream from your favorite blob field (TempBlob or some other blob) and pass it to this function.
The code is in, so it should be released with Update 15.

gunnargestsson:
Excellent solutions.  Thanks.

PR #40, opened 2017-11-28
New function and some made External

gunnargestsson:
New function that will return the mime mapping for a file name.  Path Helper functions made External.

### DIFF
--- a/BaseApp/COD419.txt
+++ b/BaseApp/COD419.txt
@@ -1,0 +1,1019 @@
+OBJECT Codeunit 419 File Management
+{
+  OBJECT-PROPERTIES
+  {
+    Date=;
+    Time=;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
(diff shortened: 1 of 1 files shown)


JesperSchulz:
Something is not quite right. I cannot see, which changes you are proposing. Under "Files Changed", it appears everything has been added, which isn't the case. COD419 is in the repo and there should hence be a limited number of changes. Could you take a look once more? Also, your changes include translations, which should be removed.
Once done, I'll take another look! Thanks!

gunnargestsson:
Well, the COD419.TXT is not in your master branch.  What will happen if you put it there?  Will you then see the changes I suggest?

JesperSchulz:
It's not? I added it earlier today! It should be in master :-)

gunnargestsson:
It is in now.  Decline this pull request, I will create a new one.

gunnargestsson:
I will ask you to add objects to master if I need them before the next pull request.

PR #42, opened 2017-11-29
Add function to get mime type from file name

gunnargestsson:
This function is needed to supply information for the http protocol.


JesperSchulz:
Your suggestion looks valid and safe. I will add it to the baseapp in the coming days.

JesperSchulz:
This pull request has been addressed through pull request #46. The code will ship with January update.

PR #43, opened 2017-12-02
New functions for Type Helper

gunnargestsson:
FormatDecimal (dot net formatting)
FormatUtcDateTime (dot net formatting)
ConvertToAscii (convert accent characters)

### DIFF
--- a/BaseApp/COD10.TXT
+++ b/BaseApp/COD10.TXT
@@ -431,6 +431,43 @@ OBJECT Codeunit 10 Type Helper
     END;
 
     [External]
+    PROCEDURE ConvertToAscii@58(NonAscii@50000 : Text) : Text;
+    VAR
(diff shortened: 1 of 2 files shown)


JesperSchulz:
Time to get some traction on this request. But first I need clarification on your ASCII function. It looks awfully much like the function we have in COD47 - StringConversionManagement. How does your implementation differ? I would hate to have two parallel implementations for the same thing. If COD47 doesn't fit your needs, maybe we should try to fix that one instead?

gunnargestsson:
You are correct.  Let's use the COD47 function and don't implement ConvertToAscii.
My bad...

JesperSchulz:
:-) OK! I'll move forward with the other two functions then! Sorry it took me so long...

JesperSchulz:
Request addressed though commit a299dbcbe828dbc90f24299bb94382d49d43da13. If you don't agree with the solution, please re-open pull request. But I think it should be alright :-)

PR #52, opened 2018-01-12
Create Image Processing Codeunit

ivandjordjevic:
Created codeunit to work with System.Drawing.Bitmap .NET library. 
We would need a way to compress images that are uploaded by users and maybe generate thumbnail pictures in some cases using .Net libraries.

Our concern is that image files in Blob fields or in new Media field type would impact performance and increase database size. Current size limit of an image that can be uploaded is 2gb (based on limit for blob field)?

Please review this codeunit that for now can:

- ResizeImage with public function: ResizeImage(VAR TempBlob : Record TempBlob;NewWidth : Integer;NewHeigh : integer;PreserveAspectRatio : Boolean);


JesperSchulz:
Thank you for your contribution! This is an interesting proposal which I'd be happy to add to our library. However, I would need tests for the code, before I can move forward. Could you add a few tests, which verify that the code does what it is supposed to do? You can find examples of tests in the Test folder of this repo. Thank you very much!

JesperSchulz:
Is there any chance you could provide the missing tests? I would love to add this functionality to our libraries and close this pull request, but I cannot move forward with this request until our test code coverage requirements are fulfilled. Kindly update the pull request or let me know, if you aren't interested in going forward with it after all.
Have a lovely weekend.

JesperSchulz:
Closing request due to missing tests. Please reopen, once tests can be provided.

PR #54, opened 2018-01-26
SignedXml support

anzwdev:
Support for signing and validating signature of signed xml documents using SignedXml .net class. Code is based on msdn examples available here: 

https://msdn.microsoft.com/en-us/library/system.security.cryptography.xml.signedxml(v=vs.110).aspx

### DIFF
--- a/BaseApp/COD50015.txt
+++ b/BaseApp/COD50015.txt
@@ -1,0 +1,137 @@
+OBJECT Codeunit 50015 Signed Xml Mgt.
+{
+  OBJECT-PROPERTIES
+  {
+    Date=26/01/18;
+    Time=10:49:58;
+    Modified=Yes;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+
+    [External]
+    PROCEDURE CheckXmlStreamSignature@1000000000(XmlStream@1000000000 : InStream;PublicKey@1000000001 : Text) : Boolean;
+    VAR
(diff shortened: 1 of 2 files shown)


JesperSchulz:
Is this a V1 blocker? Either way, I will look into it soon, but it would be good to know in order to get my prioritization straight.

anzwdev:
I would call it "soft" v1 blocker. We are using it in our V1 extension to check signature of our license xml, so it would be good to have it, but I can comment it out temporary and enable it and resubmit our v2 extension later, when you are ready.

JesperSchulz:
Ok, I will try to get it into the product by EOW then. It might spill over to the beginning of next week though, if high priority blockers or other tasks come in. But the code looks solid, so I'll def. aim for a short turn-around. Thanks once more for another valuable contribution! Much appreciated!

JesperSchulz:
I've completed the work on the Signed XML Mgt. codeunit - which wasn't much of a deal given the great work you've provided.
Please do verify if everything still works as per your expectations and if not, you know the drill ;-)
Thank you for your contribution!

Here is the commit: d5738c75f373f5f18499cc4631611502b127f409

anzwdev:
Everything works great, thank you.

JesperSchulz:
@anzwdev, an issue with your library was reported: https://github.com/Microsoft/cal-open-library/issues/73. Could you maybe take a look?

anzwdev:
I'm going to take a look at it today.

JesperSchulz:
Great! Thanks a lot!

PR #56, opened 2018-02-27
New Objects for executing Reports with RequestPages from External

FSharpCSharp:
This point refers to Issue [546](https://github.com/Microsoft/AL/issues/546) of Microsoft/AL. In it the request was introduced to make it possible to read out the options and data items of the request page via the object metadata table. Unfortunately, the NAV standard does not allow to do this. And retrieval of the Request XML document is only possible when the client is started and the RequestPage is displayed to the user. Furthermore, a lot of important information is missing to prepare this XML document accordingly. This customization is required to start a report and request page from an external Web service, and to obtain the required request parameters in order to be able to build up the XML required by NAV. The generated XML has the same structure as the query XML file, which can be passed to the REPORT.SAVE function. Only the missing options such as data type and additional information have been added.

### DIFF
--- a/BaseApp/COD50000.txt
+++ b/BaseApp/COD50000.txt
@@ -1,0 +1,276 @@
+OBJECT Codeunit 50000 Report Helper
+{
+  OBJECT-PROPERTIES
+  {
+    Date=27.02.18;
+    Time=09:44:37;
+    Modified=Yes;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+
+    PROCEDURE GetReportRequestPageStructureAsXml@6002002(ReportID@6003000 : Integer;LanguageID@6003001 : Integer;VAR ReturnValue@6003002 : Text);
+    VAR
+      Field@6004000 : Record 2000000041;
+      AllObjWithCaption@1109100008 : Record 2000000058;
+      Caption@6004006 : Text;
+      DataType@6004005 : Text;
+      Id@1109100006 : Text;
+      Name@1109100005 : Text;
+      TableCap@6004007 : Text;
+      FieldNumber@6004003 : Integer;
+      I@6004001 : Integer;
+      TableNo@6004004 : Integer;
+      DataItemList@6004008 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlNodeList";
+      InputParam@6004009 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlElement";
+      InputParameters@6004010 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlNodeList";
+      ResultXmlDocument@1109100000 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlDocument";
+      XmlAttribute@1109100003 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlAttribute";
+      XmlDocument@6004011 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlDocument";
+      XmlElement@1109100001 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlElement";
+      XmlElement2@1109100007 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlElement";
+      XmlElement3@1109100002 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlElement";
+      XmlExpression@6004012 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlElement";
+      XmlNamespaceManager@6004013 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlNamespaceManager";
+      XmlNode@1109100004 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlNode";
+    BEGIN
+      GetObjectMetadata(ReportID,XmlDocument);
+
+      IF ISNULL(XmlNamespaceManager) THEN BEGIN
+        XmlNamespaceManager := XmlNamespaceManager.XmlNamespaceManager(XmlDocument.NameTable);
+        XmlNamespaceManager.AddNamespace('xsi', 'http://www.w3.org/2001/XMLSchema-instance');
+        XmlNamespaceManager.AddNamespace('xsd', 'http://www.w3.org/2001/XMLSchema');
+        XmlNamespaceManager.AddNamespace('ns1', 'urn:schemas-microsoft-com:dynamics:NAV:MetaObjects');
+      END;
+
+      XmlNode := XmlDocument.SelectSingleNode('/Report/ID');
+      Id := XmlNode.InnerText;
+
+      XmlNode := XmlDocument.SelectSingleNode('/Report/Name');
+      Name := XmlNode.InnerText;
+
+      ResultXmlDocument := ResultXmlDocument.XmlDocument;
+      XmlElement := ResultXmlDocument.CreateElement('ReportParameters');
+
+      XmlAttribute := ResultXmlDocument.CreateAttribute('name');
+      XmlAttribute.Value := Name;
+      XmlElement.Attributes.Append(XmlAttribute);
+
+      XmlAttribute := ResultXmlDocument.CreateAttribute('id');
+      XmlAttribute.Value := Id;
+      XmlElement.Attributes.Append(XmlAttribute);
+
+      ResultXmlDocument.AppendChild(XmlElement);
+
+      XmlElement2 := ResultXmlDocument.CreateElement('Options');
+      XmlElement.AppendChild(XmlElement2);
+
+      InputParameters := XmlDocument.SelectNodes('//ns1:Controls[@xsi:type=''ControlDefinition'']',XmlNamespaceManager);
+
+      IF InputParameters.Count <> 0 THEN
+        FOR I := 0 TO InputParameters.Count-1 DO BEGIN
+          InputParam := InputParameters.Item(I);
+          XmlElement3 := ResultXmlDocument.CreateElement('Field');
+          XmlElement2.AppendChild(XmlElement3);
+
+          XmlAttribute := ResultXmlDocument.CreateAttribute('name');
+          XmlExpression := XmlDocument.SelectSingleNode(STRSUBSTNO('//ns1:Expression[@Name=''%1'']',InputParam.GetAttribute('DataColumnName')),XmlNamespaceManager);
+          XmlAttribute.Value := XmlExpression.GetAttribute('SourceExpression');
+
+          XmlElement3.Attributes.Append(XmlAttribute);
+
+          XmlAttribute := ResultXmlDocument.CreateAttribute('datatype');
+
+          DataType := XmlExpression.GetAttribute('Datatype');
+          IF DataType = '' THEN BEGIN
+            EVALUATE(TableNo,XmlExpression.GetAttribute('TableNo'));
+            EVALUATE(FieldNumber,XmlExpression.GetAttribute('FieldNo'));
+            IF Field.GET(TableNo,FieldNumber) THEN
+              DataType := FORMAT(Field.Type);
+          END;
+
+          XmlAttribute.Value := DataType;
+          XmlElement3.Attributes.Append(XmlAttribute);
+
+          XmlAttribute := ResultXmlDocument.CreateAttribute('caption');
+          XmlAttribute.Value := GetTextFromTextML(InputParam.GetAttribute('CaptionML'),LanguageID);
+          XmlElement3.Attributes.Append(XmlAttribute);
+
+          IF XmlExpression.GetAttribute('OptionString') <> '' THEN BEGIN
+            XmlAttribute := ResultXmlDocument.CreateAttribute('optionstring');
+            XmlAttribute.Value := XmlExpression.GetAttribute('OptionString');
+            XmlElement3.Attributes.Append(XmlAttribute);
+
+            XmlAttribute := ResultXmlDocument.CreateAttribute('optioncaption');
+            XmlAttribute.Value := GetTextFromTextML(InputParam.GetAttribute('OptionCaptionML'),LanguageID);
+            XmlElement3.Attributes.Append(XmlAttribute);
+          END;
+        END;
+
+      XmlElement2 := ResultXmlDocument.CreateElement('DataItems');
+      XmlElement.AppendChild(XmlElement2);
+
+      DataItemList := XmlDocument.SelectNodes('//DataItem/DataItemVarName');
+      InputParameters := XmlDocument.SelectNodes('//ns1:Controls[@xsi:type=''FilterControlDefinition'']',XmlNamespaceManager);
+
+      IF InputParameters.Count <> 0 THEN BEGIN
+        FOR I := 0 TO InputParameters.Count-1 DO BEGIN
+          InputParam := InputParameters.Item(I);
+          XmlElement3 := ResultXmlDocument.CreateElement('DataItem');
+          XmlElement2.AppendChild(XmlElement3);
+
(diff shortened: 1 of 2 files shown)


JesperSchulz:
Thank you very much for your contribution and sorry about the delay in the reply to this request. We're currently discussing the appropriate design. We will get back to you, once we've reached a conclusion.

FSharpCSharp:
@JesperSchulz 
Hello again! I wanted to know about this Pull Request? What are the chances that the functionality will be taken over? For the implementation of our solution we still need a way to query the structure of RequestPage in detail. Especially in the new cloud world there is no chance to get this information. If there is any other solution, please contact me. Unfortunately, however, no one is coming forward and I am beginning to suspect that this point has been forgotten. Please contact us and give us some feedback about this point! Many thanks in advance!

JesperSchulz:
You have not been forgotten, but I am receiving push back on solving this problem through the application for the following reasons:
- The provided solution will not include objects added through extensions
- The provided solution exposes raw metadata XML. External developers will be forced to take a dependency on our internal format, thereby locking this format down.

A request to our platform team has been created to solve this the right way, but so far I haven't heard back. With the recent Business Central release, all teams have been quite busy wrapping things up and planning the next release. I will try to push for an answer on how we proceed and when you can expect the functionality to be added. Sorry again for the delay on this one, but we would like to get this right and not rush into creating a wrong solution, which we (and you) later will have to redo. Thanks for your patience on this particular issue.

FSharpCSharp:
@JesperSchulz 
I'm well aware that you're not comfortable with someone accessing internal system tables. It was also not our intention to do so. Rather, we were looking for a way to get the information we needed. Of course, from my point of view an object model is a lot better here, and also better to wait for you. But it would have been nice, in fact, if you could have explained this topic in more detail. The subject has not only been known since yesterday, but for several months already. In my opinion, this has little to do with the pressure to publish Dynamics 365 BC. Please watch your communication. It's better to communicate a little more than not reacting to the topic properly for months.

JesperSchulz:
My apologies again. I will make sure to communicate progress (or the lack of such) in a more timely fashion going forward. I am currently waiting for a reply on my platform request and hope for good news. Once I hear back from them, I will let you know.

StanislawStempin:
@FSharpCSharp - we have now had a chance to do more investigation and planning for this feature. We will have platform support for your scenario ready in the Business Central Fall release. However, if you need a solution faster we are ready to accept your C/AL solution with a few changes. If you do not expose XML metadata externally but instead return the required information for the Request Pages through a temp table passed as a var into your function. This would be aligned with a future platform solution and not expose and add dependencies directly on our XML metadata format. Please let us know if you would like to proceed this way and we can work out the exact details of the solution together. My apologies that the process took so long and thank you for your patience so far.

FSharpCSharp:
@StanislawStempin Thank you very much for the quick information! If that's how you write it, I really think it's the best thing we wait till fall. You wrote in the last post that we could work together on the topic to meet the requirements. That sounds really good! If the whole thing works with a virtual table, then I would ask you to remember that there is a limit of 250 characters for texts here. I hope it doesn't cut off any significant information. Will you come back to me when you have talked about the concrete adjustments in the team and the actual implementation begins? We are a member of Microsoft's "Ready" Partner Program with our company, so we should receive the changes very quickly. The last AppSource webcast mentioned that there is a new build release every day.

FSharpCSharp:
@StanislawStempin 
Hello again. I have now given some thought to the implementation of this requirement, and I have also come to the conclusion that a system table is probably the best for this. I came up with the following columns for the table:

Primary Key | Column Name | DataType | Length | Comment
---------|---------|----------|---------|----------
 Yes | Report No. | Integer | | Specifies the report number.
 Yes | Type | Option (Field,DataItem) | | Specifies which option it is, whether Field or DataItem.
 Yes | Entry No. | Integer | | Is used as an internal enumeration for the current field or the current DataItem, so that the sequence of the fields or DataItems can be mapped.
 No | Name | Text | 250 | If field, this is the name of the control, otherwise it is the name of the DataItem.
 No | Data Type | Option | | Specifies the data type of the field. Should be in a similar form as in the system table "Field".
 No | Caption | Text | 250 | | Specifies the caption of the field or DataItem, in the current language, which can be changed using the "GLOBALLANGUAGE" function. Caution! There is a risk that the text will be truncated, possibly the caption should be spread over 4 fields of 250 characters each.
 No | Visible | Boolean | | Specifies whether the field or DataItem is displayed to the user.
 No | Request Filter Fields | Text | 250 | Specifies the "RequestFilterFields" of the DataItems. This is a comma-separated enumeration list with the field numbers that the user can choose from by default:  1,3
 No | Table No. | Integer | | Specifies the table number of the DataItem.

The table should always be comprehensive, i.e. also contain objects of extensions and, of course, always insist on the current metadata if possible. There is always a risk of overflow in the texts. Under certain circumstances, the text should be divided into 4 x 250 characters so that no information is lost.

FSharpCSharp:
@StanislawStempin 
Any news on this yet? I'm afraid I haven't heard from you since I described the introduction to the new table. Do you think that you can provide me with such a table until autumn, or should it run via an object model, similar to RecRef and FieldRef, because that would also have a certain charm. Please get in touch with me and let me know the current status. My boss is already asking what's wrong with this point. Thank you for your support!

StanislawStempin:
Hi @FSharpCSharp  - we have settled on a design that will combine Page Controls and Request Page Controls into one new virtual table. We will take into account the field proposals that you listed in your previous post.

FSharpCSharp:
@StanislawStempin 
And another month has passed. My boss is asking if there is any new information on this point yet. Can we expect the virtual table to be available in the coming weeks so that we can convert our product? I don't want to get on my nerves, but my boss just wants to get a statement on how things are going with the point. I once said that I would ask again and give him a rough time frame if we have met the necessary requirements for our extension.

FSharpCSharp:
@StanislawStempin 
I have now discovered in the current version that the virtual tables you mentioned already exist for pages. There is one that contains all page actions, and then the table fields. Did I misunderstand that then. I always assumed that there should be a combined table for all of this. Unfortunately there is no trace of the report page so far.

FSharpCSharp:
@StanislawStempin
In the last "Ready to Go Office Hours" it was mentioned that Microsoft plans major changes to the complete NAV reports in the future. Is that why this issue has come to a standstill? I do not expect much in principle, but a very brief answer would be appropriate here. I'm slowly getting into trouble explaining myself to my boss, because I can't catch up with anything at the moment. Oh please please give me something new on this topic, so that I can show something to my boss. That can't be that hard to get a current status here.

StanislawStempin:
Hi @FSharpCSharp,  My apologies for the delayed reply. This is still work in progress with planned completion by the Fall release of Business Central. I will update this thread once we have the feature available in the Developer Preview builds

dennisfrostlander:
Hi @FSharpCSharp,

Just wanted to share with you that we are considering a slightly modified design of the new virtual tables.

Instead of combining the data items and request page controls in the same table, we would like to split it in two as follows:
- All Control Fields. Will contain all controls for pages, report request pages, and xmlport request pages.
- Report Data Items

**All Control Fields**

Primary   Key | Column   Name | Data   Type
-- | -- | --
Yes | Object   Type | Option   (Page, Report, XmlPort)
Yes | Object   ID | Integer
Yes | Control   ID | Integer
  | Name | Text
  | Data   Type | Option (same as in Field table)
  | Option   String | Text
  | Option   Caption | Text
  | Caption | Text
  | Visible | Boolean
  | Enabled | Boolean
  | Editable | Boolean
  | Related   Table ID | Integer
  | Related   Field ID | Integer
  | Source   Expression | Text

**Report Data Items**

Primary   Key | Column   Name | Data   Type
-- | -- | --
Yes | Report   ID | Integer
Yes | Data   Item ID | Integer
  | Name | Text
  | Request   Filter Fields | Text
  | Related   Table ID | Integer
  | Data   Item Table View | Text

FSharpCSharp:
@dennisfrostlander 
That's really great news! To be honest, I also think the separation is better, because controls and DataItems are two completely different things. The only thing that struck me now is what would be missing with the DataItems the indentation level. This would be very practical, and a Boolean indicator as to whether this DataItem would be filterable on the RequestPage. Oh, and another good thing is that the caption is also available on the DataItem, because you can overwrite the display name with it. Otherwise there is too often an "integer" instead of a real display name.

dennisfrostlander:
Thanks, I will look into the additional Data Items fields.

dennisfrostlander:
Hi @FSharpCSharp, we have one problem.

Fields Visible/Enabled/Editable were specified as Boolean in the tables above, but these fields actually can contain expressions.
For example, in the existing table _Page Control Field_, these fields are exposed as strings.

First question is out of curiosity - what would you do in your code if property Visible is set to an expression?

The second problem is that if we include Option String (1024), Option Caption (1024), Visible/Enabled/Editable (512 each), and Source Expression (512) into a single table, we exceed 8000-byte record limit (by 572 bytes).

So we need to sacrifice something. Since you only requested Visible field, we could sacrifice Enabled/Editable, but I am still curious how you would handle expressions in there... :)

FSharpCSharp:
@dennisfrostlander 
These are all very valid questions. I didn't know that table size restrictions also applied to purely virtual tables. I always thought these were just programming tools to access data that wasn't stored directly in the database. I am therefore somewhat surprised that this limitation also applies here.

There seems to have been a misunderstanding at Visible. In my original request (see also the files in the pull request), it was intended that this is only valid for DataItems, because they are shown or hidden. In fact, I had not intended this property for fields at all (probably for the reasons mentioned). From my point of view, I don't need it here either. Could you look into the attached files to see which fields I get from where? Analogously I would need a corresponding illustration. Thank you very much, and have a nice weekend!

dennisfrostlander:
Oh, right, excellent. Then for now we can drop Visible/Enabled/Editable fields from the _All Control Fields_ table and solve this problem only for data items.
Regarding the limit, it exists on several layers. Even though the table is virtual, the records still need to fit into the allocated memory buffers, etc.

FSharpCSharp:
@dennisfrostlander 
Thanks for the updates! Just a short question about the availability of this table. My boss asks if we will have access to this table in one of the coming daily booths, so we can use the new table for our development? I only need to know when this is roughly available so that I don't have to download a new version every day to check whether everything is to our satisfaction.

dennisfrostlander:
Hi @FSharpCSharp,
This task is currently in the execution mode, it can take a week or two before all validation procedures are finalized. 
Once the code is committed, the functionality will appear in the bcinsider build (daily build). 
No need to check the builds every day, I will keep you posted here :)

dennisfrostlander:
Hi @FSharpCSharp,
One more question. I have compared the XML parameters that you are generating vs XML parameters that Report.RunRequestPage is generating.
There is a couple of differences in DataItem elements. You seem to add extra attributes such as "visible", "table", "requestfilterfields", etc. Additionally, you don't put any DataItemTableView expressions (VERSION(1) SORTING(Field1,Field3), etc).

Is this done on purpose? Could you please give more details on how exactly you are modifying the resulting XML in the external service and what is the final XML that you are sending back to NAV to execute the report?

Just want to make sure I understand the whole flow in details.

dennisfrostlander:
Ok, so extending the XML file format is not needed, that's clear.
But do I understand the full flow that you are building correctly here?

The initial discussion started with https://github.com/Microsoft/AL/issues/546:
```
If you want to run a report outside of Dynamics NAV and pass parameters...
```

So the bottom line is that we want generate the ReportParametersXml from some metadata, without calling Report.RunRequestPage. 
The old solution was doing that using `Object Metadata` directly, which was not desired, it was missing reports in extensions and so on.
The new solution helps achieve the same thing (build ReportParametersXml) with two new virtual tables as discussed above.

Then, as far as I understand, in your end-to-end user flow, you would create ReportParametersXml in NAV, send it out to some external system where the user would populate Field values, then send the filled-in xml file back to NAV and run the report with these parameters.
Is that correct?

FSharpCSharp:
@dennisfrostlander 
Yes, exactly, here again the whole story step by step. We use our programming to call up reports from NAV using web services. We want to make it as pleasant as possible for the user, and therefore we offer the luxury of calling up existing reports with a slimmed-down version of the request page. He can then save the reports in any available format, and in a very convenient way. Unfortunately, we had to go through the metadata for this, because the standard did not help us in any way. The first problem was that the request XML, which you get only comes back if you have a GUI. Furthermore, the information in it was very sparse. That is, we could not work on that. This is why the parameters Visible and RequestFilterFields are also displayed on the DataItem. We simulate the whole system so that the user does not have to use two different systems. That's how he finds himself. 

That's why I opened this pull request to give people the opportunity to get at least a bored version of the request XML. The handover to the report went very well because we had only used arguments for it. This should only serve as a temporary solution, and was in no way our wish. For this reason we do not need the Enable and Visible properties for the data fields, because the NAV standard always passes all fields in this XML. Furthermore, there are rarely reports that are so complex that they play around a lot. Up to now, no customer has complained to us about missing functionalities. 

I hope this will help to clarify matters a little bit. I think it's great that you're questioning all this so we can work out the best solution. I really appreciate this!

dennisfrostlander:
Hi @FSharpCSharp,

The bcinsider build has now the new virtual tables to try out.
But before you continue, an **important notice**. In one of the next builds, the virtual tables will change their IDs (we had to do it because of some licensing issues). So the code you try today may stop working, but you can simply recompile the code on newer System symbols and everything should continue working as before.

I have tested this on bcinsider build _W1 13.0 (24012)_, which I got by simply running `docker run -e accept_eula=Y bcinsider.azurecr.io/bcsandbox-master`

The two new tables are called "All Control Fields" and "Report Data Items".
You can use the following codeunit to generate the ParametersXml document.
There are also some additional columns (e.g. "Sorting Fields"), that should help you generate the extended XML document as you have shared before.

```JavaScript
codeunit 50103 "Report Xml Parameters Builder"
{
    procedure "Get Report Page Structure As Xml"("Report ID": Integer): XmlDocument
    var
        "Report Metadata": Record "Report Metadata";
        "Xml Parameters": XmlElement;
        "Result": XmlDocument;
    begin
        "Report Metadata".SetRange(ID, "Report ID");
        "Report Metadata".Find('-');

        "Xml Parameters" := XmlElement.Create('ReportParameters');
        "Xml Parameters".SetAttribute('name', "Report Metadata".Caption);
        "Xml Parameters".SetAttribute('id', Format("Report ID"));

        "Xml Parameters".Add("Create Xml Options"("Report ID"));
        "Xml Parameters".Add("Create Xml Data Items"("Report ID"));

        Result := XmlDocument.Create();
        Result.Add("Xml Parameters");

        exit(Result);
    end;

    local procedure "Create Xml Options"("Report ID": Integer): XmlElement
    var
        "All Control Fields": Record "All Control Fields";
        "Xml Options": XmlElement;
        "Xml Field": XmlElement;
    begin
        "All Control Fields".SetRange("Object Type", "All Control Fields"."Object Type"::Report);
        "All Control Fields".SetRange("Object ID", "Report ID");
        "All Control Fields".Find('-');

        "Xml Options" := XmlElement.Create('Options');

        repeat
            "Xml Field" := XmlElement.Create('Field');
            "Xml Field".SetAttribute('name', "All Control Fields"."Source Expression");

            "Xml Options".Add("Xml Field");
        until "All Control Fields".Next() = 0;

        exit("Xml Options");
    end;

    local procedure "Create Xml Data Items"("Report ID": Integer): XmlElement
    var
        "Report Data Items": Record "Report Data Items";
        "Xml Data Items": XmlElement;
        "Xml Data Item": XmlElement;
    begin
        "Report Data Items".SetRange("Report ID", "Report ID");
        "Report Data Items".Find('-');

        "Xml Data Items" := XmlElement.Create('DataItems');

        repeat
            "Xml Data Item" := XmlElement.Create('DataItem');
            "Xml Data Item".SetAttribute('name', "Report Data Items".Name);
            "Xml Data Item".Add(XmlText.Create("Report Data Items"."Data Item Table View"));

            "Xml Data Items".Add("Xml Data Item");
        until "Report Data Items".Next() = 0;

        exit("Xml Data Items");
    end;
}
```

A simple test:

```JavaScript
pageextension 50100 CustomerListExt extends "Customer List"
{
    trigger OnOpenPage();
    var
        Builder: Codeunit "Report Xml Parameters Builder";
        ParametersXml: XmlDocument;
        ResultFile: File;
        ResultStream: OutStream;
        ParametersString: Text;
    begin
        // 1. Generate Parameters XML
        ParametersXml := Builder."Get Report Page Structure As Xml"(205);

        // 2. Update the values in the external service
        "Update Field Value"(ParametersXml, 'NoOfCopies', '0');
        "Update Field Value"(ParametersXml, 'ShowInternalInfo', 'false');
        "Update Field Value"(ParametersXml, 'ArchiveDocument', 'false');
        "Update Field Value"(ParametersXml, 'LogInteraction', 'true');
        "Update Field Value"(ParametersXml, 'DisplayAssemblyInformation', 'false');

        // 3. Execute report in NAV
        ParametersXml.WriteTo(ParametersString);

        Message(ParametersString);
    end;

    local procedure "Update Field Value"(var Document: XmlDocument; "Field Name": Text; "Field Value": Text)
    var
        Node: XmlNode;
    begin
        Document.SelectSingleNode('//Field[@name="' + "Field Name" + '"]', Node);
        Node.AsXmlElement().Add(XmlText.Create("Field Value"));
    end;
}
```

FSharpCSharp:
@dennisfrostlander 
As promised, here is a short statement on the new tables at the end of the week. I have now looked at them in detail and I have to say that they really do contain everything I had in mind. Great work! This means that we can now implement our requirement as it originally was. It's really great to see how well the collaboration with Microsoft works in the new Business Central world. Please pass on a big praise to the whole team. The last few months have been very productive and innovative. You can really feel the joy of working at Business Central in the employees. Please keep up the good work! It could never have been better.

dennisfrostlander:
Since we have an alternative solution, let's close this pull request.

PR #58, opened 2018-04-02
New TryDownloadFromUrl function for TempBlob.

gunnargestsson:
Need an easy way to download any file directly into TempBlob


JesperSchulz:
Looks fine to me. Will start processing your request.

PR #61, opened 2018-04-10
Soap Proxy Client

gunnargestsson:
With this code we will be able to use most of the soap web services out there directly from AL without having to go though Azure Funztions or to manually create the soap request.


gunnargestsson:
Hi Jesper.
We need this function as well

    [External]
    PROCEDURE SetSecurityProtocolType@38(SecurityProtocolTypeText@1000 : Text);
    VAR
      ServicePointManager@1001 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Net.ServicePointManager";
      SecurityProtocolType@1002 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Net.SecurityProtocolType";
    BEGIN
      ParseEnum(SecurityProtocolType,GETDOTNETTYPE(SecurityProtocolType),SecurityProtocolTypeText);
      ServicePointManager.SecurityProtocol := SecurityProtocolType;
    END;


to be able to set

[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12

JesperSchulz:
ok! Hope to get around to look at your request beginning of next week.

gunnargestsson:
We might even need two more functions to set the GlobalParameter from Xml and get the GlobalParameter from Xml.  Will create those..

JesperSchulz:
I will hold back until you're done :-) Let me know when you are!

gunnargestsson:
Ok.  I am done and will be writing a blog about this with more usage examples.

    [External]
    PROCEDURE SetGlobalParameterValueFromXml@41(TempBlob@1000 : Record 99008535);
    VAR
      XmlSerializer@1001 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.Serialization.XmlSerializer";
      XmlDoc@1002 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlDocument";
      XmlNodeReader@1003 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlNodeReader";
      InStr@1004 : InStream;
    BEGIN
      TempBlob.Blob.CREATEINSTREAM(InStr);
      XmlDoc := XmlDoc.XmlDocument;
      XmlDoc.Load(InStr);
      XmlNodeReader := XmlNodeReader.XmlNodeReader(XmlDoc.DocumentElement);
      XmlSerializer := XmlSerializer.XmlSerializer(GlobalParameter.GetType,GetTargetNamespace);
      GlobalParameter := XmlSerializer.Deserialize(XmlNodeReader);
    END;


    [External]
    PROCEDURE GetGlobalParameterAsXml@39(VAR TempBlob@1000 : Record 99008535);
    VAR
      XmlSerializer@1001 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.Serialization.XmlSerializer";
      OutStr@1002 : OutStream;
    BEGIN
      TempBlob.INIT;
      TempBlob.Blob.CREATEOUTSTREAM(OutStr);
      XmlSerializer := XmlSerializer.XmlSerializer(GlobalParameter.GetType,GetTargetNamespace);
      XmlSerializer.Serialize(OutStr,GlobalParameter);
    END;

gunnargestsson:
One question though.  If I want import a X509 Certificate to store in my BLOB field.  Can I do that from AL or do I need to add that function to this Codeunit as well ?

Perhaps a function in TempBlob to import a Certificate would be nice ?

JesperSchulz:
I don't recall us having anything that meets your needs, no! I like your idea of adding a little helper for that to TempBlob :) Or do we have a certificate management codeunit of some sort? It's doesn really fit into the encryption management one, does it? Your call :-)

gunnargestsson:
What function can we use from AL to import a file into TempBlob ?

gunnargestsson:
Also, I suggest that you add field 3, File Name Text[250] to TempBlob

kine:
Importing of certificate into Blob is done in Czech localization I think with use of the encryption, and I will expect to encrypt the certificate inside the DB... ;-)

gunnargestsson:
Hi Jesper.
I have updated the code and added one more codeunit for X509 Certificates.
Ready for you to work on.

JesperSchulz:
Hi Gunnar,

let us schedule a call to discuss the format of pull requests we'd ideally like to receive in this repository. We need to have focus on reusability when wrapping .NET methods. I will reach out to you today to set something up.

gunnargestsson:
Now that BC for Iceland is soon online we need this codeunits, at least in the Icelandic Localization.  Without these codeunits will not be able to communicate with tax authorities, pension funds and online banking.

JesperSchulz:
Hi Gunnar,

I will have to discuss your solution once more with my colleagues. As discussed previously, you are not following the format we'd like to use going forward, which is why I'm receiving push back on your pull request(s). Your library will fit your specific needs, rather than introducing reusable .NET wrappers as found in codeunits 3000 and upwards. I do, however, see the usefulness of your library and understand why you need it. Let me get back to you beginning of next week. I need to play around with a few solutions. Maybe we can meet in the middle :-)

gunnargestsson:
Hi

Jesper

This is a generic solution.  I believe the every AL developer can use this library to connect to any SOAP service out there.

The other codeunit that allows you to import and save a certificate is as generic as the can get.


Kind regards,
Gunnar Þór Gestsson

JesperSchulz:
Sure, in _that_ sense it is reusable, but if someone else wanted to use e.g. System.Security.Cryptography or System.Convert, they wouldn't be able to do that with your solution. They'd have to submit another PR, which is what we'd like to avoid. By wrapping the .NET libraries like in COD3000, the wrappers become multi purpose and resuable and every partner can create their own libraries using the wrappers without having to submit them all to the baseapp. We would be able to isolate and reuse .NET wrappers. Hence the push back. We have to start isolating .NET calls to a higher degree, as we else won't be able to cope with the inflow of requests.

gunnargestsson:
Agreed. In that sense this is not a dotnet wrapper but rather a dotnet library.

We should be able to move some of the functions to wrappers. Do you want me to take a look at the proxy from that point?

On Jun 22, 2018 08:01, Jesper Schulz-Wedde <notifications@github.com> wrote:

Sure, in that sense it is reusable, but if someone else wanted to use e.g. System.Security.Cryptography or System.Convert, they wouldn't be able to do that with your solution. They'd have to submit another PR, which is what we'd like to avoid. By wrapping the .NET libraries like in COD3000, the wrappers become multi purpose and resuable and every partner can create their own libraries using the wrappers without having to submit them all to the baseapp. We would be able to isolate and reuse .NET wrappers. Hence the push back. We have to start isolating .NET calls to a higher degree, as we else won't be able to cope with the inflow of requests.

JesperSchulz:
If you could wrap, that would be fantastic! And then the push back would be gone and I could submit your code. The problem is, that all code we accept now which isn't using wrappers, adds work load to our own backlog as we will have to refactor it soon. If you could do so, it would be highly appreciated!

JesperSchulz:
Cancelling pull request upon Gunnars request. He will create a new PR for the individual .NET wrappers to make the solution better reusable.

PR #62, opened 2018-04-16
New function to read directly from TempBlob and other External

gunnargestsson:


gunnargestsson:
I would also suggest a similar read function in TAB1235.

PR #64, opened 2018-04-23
Enable Xml values bigger than 250char

gunnargestsson:
Remove the length limitation of Xml node value


JesperSchulz:
Lovely. I will look into your code this week! Thanks, Gunnar!

JesperSchulz:
Code does not compile. You are missing declaration of new field "Value Blob". Is that all that is missing, or were other parts left out? Please double check, before I proceed.

PR #65, opened 2018-04-24
Json Interface Management for Interface Codeunits

gunnargestsson:
I like to use TempBlob with interface codeunits.  This code helps working with the interface json.


JesperSchulz:
Hi Gunnar,

let us schedule a call to discuss the format of pull requests we'd ideally like to receive in this repository. We need to have focus on reusability when wrapping .NET methods. I will reach out to you today to set something up.

JesperSchulz:
Cancelling pull request upon Gunnars request. He will create a new PR for the individual .NET wrappers to make the solution better reusable.

PR #67, opened 2018-05-04
DotNet Uri wrapper

freddydk:
EscapeDataString and UnescapeDataString for creating Urls safely

### DIFF
--- a/BaseApp/COD3023.txt
+++ b/BaseApp/COD3023.txt
@@ -1,0 +1,39 @@
+OBJECT Codeunit 3023 DotNet_Uri
+{
+  OBJECT-PROPERTIES
+  {
+    Date=04-05-18;
+    Time=09:10:02;
+    Modified=Yes;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      DotNetUri@1000 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Uri";
+
+    [External]
+    PROCEDURE EscapeDataString@1(Text@1000 : Text) : Text;
+    VAR
+      DotNetArray@1001 : DotNet "'mscorlib'.System.Array";
+    BEGIN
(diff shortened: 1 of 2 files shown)


BaseApp/COD3023.txt
```diff
@@ -0,0 +1,39 @@
+OBJECT Codeunit 3023 DotNet_Uri
+{
+  OBJECT-PROPERTIES
+  {
+    Date=04-05-18;
```
JesperSchulz: Date, Time and Modified flag should be cleared.

BaseApp/COD3023.txt
```diff
@@ -0,0 +1,39 @@
...
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      DotNetUri@1000 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Uri";
```
JesperSchulz: No necessary to specify a version. But it does not harm as such. But this would suffice:
DotNetUri@1000 : DotNet "'System'.System.Uri";

Test/COD145023.txt
```diff
@@ -0,0 +1,39 @@
+OBJECT Codeunit 145023 Test_DotNet_Uri
+{
+  OBJECT-PROPERTIES
+  {
+    Date=04-05-18;
```
JesperSchulz: Again, please clear date, time and modified (remove the latter).

Test/COD145023.txt
```diff
@@ -0,0 +1,39 @@
...
+  {
+    Subtype=Test;
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      Assert@1170000000 : Codeunit 130000;
```
JesperSchulz: Looks like the wrong ID range was used here ;-)

BaseApp/COD3023.txt
freddydk: Actually, in all other dotnet wrapper codeunits using system.dll, the version is specified as well. It doesn't work without.

PR #76, opened 2018-07-06
Dotnet string split

gunnargestsson:


JesperSchulz:
Looking good, but as always, I need tests before I can process your PR :-) Please supply.

PR #79, opened 2018-07-08
Dot net certificate

gunnargestsson:
I think I got the picture...


gunnargestsson:
Are we sure thet the CopyToInstream function in COD3009 is not CopyFromInstrem ???

BaseApp/COD3009.TXT
```diff
@@ -32,6 +32,13 @@ OBJECT Codeunit 3009 DotNet_MemoryStream
       DotNetMemoryStream := DotNetMemoryStream.MemoryStream(DotNetArray)
     END;
 
+    [External]
```
JesperSchulz: Why do we need this function? The same can be achieved by first calling InitMemoryStream and then CopyToInStream. We should try to avoid adding helper functions into wrappers. Maybe create a separate type helper?

BaseApp/COD146027.TXT
```diff
@@ -0,0 +1,156 @@
...
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      Assert@1000 : Codeunit 130000;
+
+    [Test]
+    [LineStart(75)]
```
JesperSchulz: What are these LineStart tags? :-D

BaseApp/COD3027.TXT
```diff
@@ -0,0 +1,72 @@
...
+    PROCEDURE ExportToArray@13(VAR DotNet_Array@1001 : Codeunit 3000);
+    VAR
+      X509ContentType@1003 : DotNet "'mscorlib'.System.Security.Cryptography.X509Certificates.X509ContentType";
+    BEGIN
+      DotNet_Array.SetArray(X509Certificate2.Export(X509ContentType.Pkcs12));
+    END;
+
+    [External]
+    [LineStart(42)]
+    PROCEDURE GetCertificateDetails@2(VAR FriendlyName@1001 : Text;VAR Thumbprint@1002 : Text;VAR Issuer@1003 : Text;VAR Subject@1000 : Text;VAR Expiration@1004 : DateTime;VAR HasPrivateKey@1008 : Boolean);
```
JesperSchulz: We should split the getters. If a new property comes along, we would need to change the signature of this getter, which is a breaking change. By having individial getters, we avoid that problem!

Setup.json
```diff
@@ -1,13 +1,13 @@
 ﻿{
```
JesperSchulz: No need to add this file in every check-in ;-) You can exclude it from your submission or add it to the ignore list in git.

gunnargestsson:
What troubles me is that the functionality does not match the function name.  CopyToInStream function is CopyFromInStream.

Kveðja/Kind regards,
Gunnar Þór Gestsson

gunnargestsson:
The LineStart now is in every export from bcmaster image.

Kveðja/Kind regards,
Gunnar Þór Gestsson

gunnargestsson:
Agree on splitting the getters.

Kveðja/Kind regards,
Gunnar Þór Gestsson

BaseApp/COD146027.TXT
esbenk: They are for integration with VsCode, so F12 can navigate to the method source in the base-app. I think we only add them when exporting to new syntax now.

JesperSchulz:
Just returned from the triage meeting, where we were discussing all requests. For this request, usage example on how this is used in the cloud was requested. This wrapper is perfectly safe on prem, but before the functions can be exposed, we need more details.

gunnargestsson:
Files updated

grobyns:
Hi Gunnar,

I received feedback that this code would not work in our SaaS environment because of the use of the LocalMachine store. In a multi-server setup, the certificate would be stored on one machine and not another and so the results would be unpredictable.

br/Gert

gunnargestsson:
That I am aware of. That is why thouse functions are not External.

Can only be used for on prem installations.

On Aug 2, 2018 8:56 AM, Gert Robyns <notifications@github.com> wrote:

Hi Gunnar,

I received feedback that this code would not work in our SaaS environment because of the use of the LocalMachine store. In a multi-server setup, the certificate would be stored on one machine and not another and so the results would be unpredictable.

br/Gert

grobyns:
Hi Gunnar,

On-prem you can also have multi-server installs. In any case, the cal open library is meant to help people move to the cloud. If this doesn't work in the cloud then I think we'll have to reject the PR.

br/Gert

gunnargestsson:
Then just take the first codeunit that handles certificates that you import into the database.

On Aug 2, 2018 9:13 AM, Gert Robyns <notifications@github.com> wrote:

Hi Gunnar,

On-prem you can also have multi-server installs. In any case, the cal open library is meant to help people move to the cloud. If this doesn't work in the cloud then I think we'll have to reject the PR.

br/Gert

PR #80, opened 2018-07-10
Create AESEncryptionManagement.txt

quadfin:
Submitting a proposal for AES symmetric encryption -- because we need a symmetric encryption algorithm to enable private data communication between an extension app and external web services.

### DIFF
--- a/BaseApp/AESEncryptionManagement.txt
+++ b/BaseApp/AESEncryptionManagement.txt
@@ -1,0 +1,710 @@
+OBJECT Codeunit 50000 AES Encryption Management
+{
+  OBJECT-PROPERTIES
+  {
+    Date=07/08/18;
+    Time=[ 4:10:55 PM];
+    Modified=Yes;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+
+            // Public Interface ======================================================================================
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      OptionKeySize@1000000001 : '256,192,128';
+      OptionCipherMode@1000000002 : 'CBC,CFB,CTS,OFB,ECB';
+      OptionPaddingMode@1000000005 : 'PKCS7,ANSIX923,ISO10126,Zeros,None';
+      CipherMode@1000000003 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Security.Cryptography.CipherMode";
+      PaddingMode@1000000006 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Security.Cryptography.PaddingMode";
+      Key@1000000008 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Array";
+      IV@1000000007 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Array";
+      HaveKey@1000000009 : Boolean;
+      HaveIV@1000000010 : Boolean;
+      SystemType@1000000004 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Type";
+
+    [External]
+    PROCEDURE SetKeySize@1000000022(Size@1000000000 : '256,192,128');
+    BEGIN
+
+      OptionKeySize := Size;
+    END;
+
+    [External]
+    PROCEDURE SetCipherMode@1000000024(Mode@1000000000 : 'CBC,CFB,CTS,OFB,ECB');
+    BEGIN
+
+      OptionCipherMode := Mode;
+    END;
+
+    [External]
+    PROCEDURE SetPaddingMode@1000000000(Mode@1000000000 : 'PKCS7,ANSIX923,ISO10126,Zeros,None');
+    BEGIN
+
+      OptionPaddingMode := Mode;
+    END;
+
+    [External]
+    PROCEDURE SetKey@1000000008(VAR TempBlob@1000000000 : Record 99008535);
+    BEGIN
+
+      IF NOT TrySetKey(TempBlob) THEN ;
+    END;
+
+    [External]
+    PROCEDURE SetIV@1000000009(VAR TempBlob@1000000000 : Record 99008535);
+    BEGIN
+
+      IF TrySetKey(TempBlob) THEN ;
+    END;
+
+    [External]
+    PROCEDURE SetPassword@1000000015(VAR TempBlob@1000000000 : Record 99008535);
+    BEGIN
+
+      IF TrySetPassword(TempBlob) THEN ;
+    END;
+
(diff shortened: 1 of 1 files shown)


JesperSchulz:
Thank you very much for your contribution! Unfortunately, we cannot accept the code in the format provided. We need .NET wrappers to be as resuable as possible, which is why we would ask you to look over your request once more. For examples, please refer to COD3000 and upwards in this repository.

Here are our best practices for creating wrappers (in a rough draft):

	- Add the dotnet assembly as a global, prefix the name with DotNet
	- There should be only 1 global
	- The signature of every function should match the signature of the corresponding .NET method
		○ Exception: Return values can be VAR parameters.
	- You cannot not take a dependency on the BaseApp
		○ i.e. The 3000 range must be compilable in a new empty database
	- Don't add helpers to the wrappers. Isolate type helper functions in codeunits of their own.
		○ Small helpers can be added in the 3750..3999 range. These should not be large functions with logic but rather small helpers that make it easier to use the wrapper functions. Logic should be placed in normal BaseApp as usual.
	- Add a Get<dotnet>
		○ This is a global function
		○ This is NEVER an external function
		○ It has 1 parameter
			§ Has the same name as the global but with a '2' appended
			§ Is a VAR
	- Add a Set<dotnet>
		○ This is a global function
		○ This is NEVER an external function
		○ It has 1 parameter
			§ Has the same name as the global but with a '2' appended
			§ Is a VAR
	- All functions in the wrapper are global
	- If a function is SAFE, it is made External
		○ The same rules apply for internal/external as for everything else
		○ The internal/external is meant to protect tenants from each other
		○ The internal/external is meant to protect our service
			§ No file access is allowed!
	- Except for the Get and Set functions, no dotnet is allowed as parameters
		○ If a function requires a dotnet parameter:
			§ the dotnet variable is added as a local
			§ The wrapper for the dotnet is added as a parameter (i.e. codeunit parameter)
			§ The dotnet variable is retrieved from the the codeunit parameter by using the Get method
		○ If a function returns a dotnet variable:
			§ The dotnet variable is added as a local
			§ The wrapper for the dotnet is added as a VAR parameter
The dotnet variable is set in the wrapper by using the Set method.

quadfin:
Hello Jesper,

 

Thank you for the response.

Can you please comment on the following?

 

In the current AL extension development environment there is a (critical) need for a robust symmetric encryption interface.  Presently, to enable secure data communication with external web services, the AL developer must role his own symmetric encryption, which leads to not very secure hacks at a solution.  The availability of a robust symmetric encryption interface would alleviate that problem.  The AES encryption algorithm is an obvious choice.

 

Unfortunately there is no single existing dotnet class that provides a complete AES encryption interface.  In dotnet, to implement AES encryption, one must amalgamate multiple dotnet interfaces to yield a solution.

 

The intent of the submitted codeunit is not to create a wrapper around one .NET class since that does not achieve the intended goal.  Instead, the intent of the submitted codeunit is to provide an AL callable interface for AES encryption.

 

The submitted codeunit exposes an interface to enable AES encryption/decryption of data for string, byte[] (blob), or stream data sources.  It also provides an interface to enable the caller to manage passwords in a secure and customized manner.  These are both necessary requirements for a symmetric encryption interface.  Unfortunately, modifying the interface per the coding guidelines identified is not practical since this functionality is not based on the single class wrapper design paradigm of those guidelines.

 

If the sole intent of the Cal-Open-Library is to provide wrappers around individual dotnet classes then this codeunit does not satisfy that criteria.

If that is the case and this codeunit (or a variant of it) will not be considered for inclusion then I ask:  Is there another way to submit this functionality for inclusion in the C/AL base?

 

I would be very open to massaging the interface to meet any identified coding practices.  

 

Further comment: In a more perfect world this type of functionality would not be distributed in codeunit form – instead it would exist as an application exposed API, similar to the existing ENCRYPT/DECRYPT API.  I would assume however that the timeframe for that form of implementation is quite long – hence this request.

 

Thanks for your time.

Regards,

Phil

FSharpCSharp:
@quadfin @JesperSchulz 
I can also say here from some projects that we need a robust encryption solution. In addition to the symmetric encryption already mentioned, however, in my view there must also be an asymmetric one. The Encryption Mangement Codeunit offered in the standard NAV and BC also offers far too little functionality in my opinion. Especially when it comes to encryption, the system is anything but transparent. From my point of view it would be very important that the standard streams also receive cryptographic methods. How am I supposed to read or write something encrypted if I have a lot of data? The storage in Microsoft's Online ServiceTier will also be limited, and therefore not finite. Nowadays it is actually state of the art to work with encryption. Even in my favorite.NET world, we all get made up again and again that everything should be encrypted, and access should be restricted as much as possible. Why is this not possible within NAV / BC?

grobyns:
Hi,

The intent of the cal open library is to temporarily unblock partners who are moving their solutions to the cloud and who are currently using .net, as that is not supported. In the long term, these wrappers should disappear as capabilities are added to the al language.
You should of course post requests for missing capabilities in the al language by posting an issue here: https://github.com/microsoft/al/issues so the team is aware and can build their backlog.

Keeping the wrappers clean following the guidelines we've set out will make the transition away from the wrappers easier. 
We know these wrappers are not always easy to use so we do accept helper functions, but we put those in another range. 
We distinguish between two cases of helpers:
- small helpers which serve just to make the use of the wrapper more accessible. These would contain business logic.
- helper functions which wrap some form of business logic

We can accept both but we'll be placing those in separate ranges of app objects. We'll point these helper methods out in code reviews when we see them of course.

I hope this clarifies our approach a little.

br/Gert

quadfin:
Gert,

Thanks for taking the time to respond & clarify.
I know how easy it is to let threads like this slip quietly away so it is great to see the effort your team puts in to address these concerns.

With respect to your design description and my submission - please consider:

- I have a NAV (C/AL) add-in that I am getting ready to release as an extension app
- I need to privately communicate data (license information) between the app and it's web portal
- I currently use .NET AES 256 based encryption to pass the data (in the C/AL implementation)
- The AL language (& NAV C/AL) has no equivalent functionality 
  Note: The NAV native ENCRYPT/DECRYPT API does not suffice as the keys are not private (to my app)

So, IMHO, my request seems to fit within the parameters of your design description..

- A need to replace .NET dependencies
- The AL language APIs are currently insufficient to the task
- Provide a minimalistic C/AL wrapper around required functionality (AES encryption)

Can I bother your team to have another look at the submitted codeunit (AES Encryption Management) and provide feedback as to the changes required to morph it into a suitable submission?
Note: Jesper did comment - but his statement was quite general and hence not very actionable (or perhaps I'm just a bit too dense to figure it out).

Thanks again.
Regards,
Phil

grobyns:
Hi again,

grobyns:
Hi,
as I understood Jesper's reply, he did not reject the PR based on what the code does but rather on the form it is currently in.
If you refactor the code following the guidelines Jesper sent, I'd be happy to take a look.
I think you would need to split the code into separate objects so that
- wrappers are pure wrappers (these would end up in the 30xx+ range
- a codeunit of small helpers, with no logic (these would end up in the 375x+ range
- a codeunit of helpers with logic (these would likely end up in a mgmt. codeunit as part of normal baseapp)

br/Gert

quadfin:
Got it.  Will do.

PR #82, opened 2018-07-10
Wrapper for XslCompiledTransform class

niku-live:
In our custom add-on we heavily use Xslt transformations to create documents from provided XML data. So for us it is important to have XslTransform wrapper. In this commit I added wrappers for System.Xml.Xsl.XslCompiledTransform and System.Xml.XmlReader. I think it is useful C/AL open library extension


BaseApp/COD50003.txt
```diff
@@ -0,0 +1,91 @@
...
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      DotNetXslTransform@17024400 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.Xsl.XslCompiledTransform";
+
+    [External]
+    PROCEDURE CreateFromBlob@17024400(VAR TempBlob@17024400 : Record 99008535);
```
JesperSchulz: Having such a function makes total sense, however, we should try to isolate helper functions from the actual wrappers. A wrapper codeunit cannot take any dependecy on the base application (e.g. TempBlob), as the BaseApplication eventually will have to take a dependency on the wrapper codeunits. Ideally, this function would go into an XML helper codeunit, much like COD10 Type Helper, but a little more structured ;-)

BaseApp/COD50003.txt
```diff
@@ -0,0 +1,91 @@
...
+    VAR
+      TransformStream@17024401 : InStream;
+    BEGIN
+      TempBlob.Blob.CREATEINSTREAM(TransformStream);
+      CreateFromStream(TransformStream);
+    END;
+
+    [External]
+    PROCEDURE CreateFromStream@17024424(XslTransformationStream@17024400 : InStream);
+    VAR
```
JesperSchulz: Again, a fine helper function. And eventhough this one does not depend on the BaseApp, we should isolate it with other helpers.

BaseApp/COD50003.txt
```diff
@@ -0,0 +1,91 @@
...
+      XmlReader@17024401 : Codeunit 50004;
+    BEGIN
+      XmlReader.CreateFromStream(XslTransformationStream);
+      CreateFromXmlReader(XmlReader);
+    END;
+
+    [External]
+    PROCEDURE CreateFromXmlReader@17024401(VAR XmlReader@17024400 : Codeunit 50004);
+    VAR
```
JesperSchulz: And a 3rd candidate for a helper!

BaseApp/COD50003.txt
```diff
@@ -0,0 +1,91 @@
...
+      DotNetXmlReader@17024401 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlReader";
+    BEGIN
+      XmlReader.GetXmlReader(DotNetXmlReader);
+      DotNetXslTransform := DotNetXslTransform.XslCompiledTransform;
+      DotNetXslTransform.Load(DotNetXmlReader);
+    END;
+
+    [External]
+    PROCEDURE TransformBlob@17024408(VAR TempBlob@17024405 : Record 99008535;VAR DestinationStream@17024404 : OutStream);
+    VAR
```
JesperSchulz: 4th

BaseApp/COD50003.txt
```diff
@@ -0,0 +1,91 @@
...
+      SourceStream@17024400 : InStream;
+    BEGIN
+      TempBlob.Blob.CREATEINSTREAM(SourceStream);
+      TransformStream(SourceStream, DestinationStream);
+    END;
+
+    [External]
+    PROCEDURE TransformStream@17024428(VAR SourceXmlStream@17024405 : InStream;VAR DestinationStream@17024404 : OutStream);
+    VAR
```
JesperSchulz: 5th

BaseApp/COD50003.txt
```diff
@@ -0,0 +1,91 @@
...
+      XmlDocument@17024406 : Codeunit 3013;
+    BEGIN
+      XmlDocument.Load(SourceXmlStream);
+      TransformXmlDocument(XmlDocument, DestinationStream);
+    END;
+
+    [External]
+    PROCEDURE TransformXmlDocument@17024407(VAR XmlDocument@17024405 : Codeunit 3013;VAR DestinationStream@17024404 : OutStream);
+    VAR
+      DotNetXPathNavigatable@17024401 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XPath.IXPathNavigable";
```
JesperSchulz: Now THIS is what we should have in the wrapper! And ideally, we should align function and method name.

BaseApp/COD50004.txt
```diff
@@ -0,0 +1,50 @@
...
+
+  }
+  CODE
+  {
+    VAR
+      DotNetXmlReader@17024401 : DotNet "'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Xml.XmlReader";
+
+    [External]
+    PROCEDURE CreateFromBlob@17024414(VAR TempBlob@17024400 : Record 99008535);
+    VAR
```
JesperSchulz: Another helper

JesperSchulz:
Btw., we'll be posting some guidelines very soon, which sum up the "problems" you ran into here. There is no way you could have known. I hope the many comments don't discourage you. We really appreciate your efforts!

BaseApp/COD50004.txt
```diff
@@ -0,0 +1,50 @@
+OBJECT Codeunit 50004 DotNet_XmlReader
```
JesperSchulz: We do have a native implementation of this in the al language. Is there any reason why you need this impementation instread? We would have to make the function here [Internal], as we don't want two competing constructs for the same purpose.

BaseApp/COD50003.txt
JesperSchulz: Actually, it would be better to use the .Load(String) method and remove the dependency on the XML Reader, as we shouldn't expose the XML Reader functions (see previous comment in XML Reader).

niku-live:
Hi,

Don't worry, it's not discouragement and more like an encouragement because I see that you look at my ideas seriously and just want to make them even better.  Four eyes is always better than two :) and I agree with all your comments.

So I tried to move all helper functions to separate codeunit. Also I removed DotNet_XmlReader codeunit, basically I only needed it as a helper. Still I cannot use .Load(string) method as that string in a method means "URI to file" and not "text value of xslt transformation", but I created "Create" function which I think is quite good solution. Waiting for your comments.

niku-live:
Hello,

I updated my PR a little bit:
1. Fixed all IDs and some procedure naming.
2. Procedure Create was split into 2 parts - constructor procedure XslCompiledTransform and Load procedure. Now it represents XmlCompiledTransform class more closely than before.
2. Removed dependency on XmlReader - XmlTransform can be loaded from XmlDocument, so I changed Load procedure - now it loads from XmlDocument not XmlReader. As XmlDocument wrapper was already included in BaseApp, this dependency should be acceptible.

I hope that now this PR meets all requirements, inform me, if you need any other changes.

JesperSchulz:
Great! Thank you :-) I'm sorry I haven't processed this one yet. I got stuck in the middle and had to turn my attention towards our upcoming Fall release... As soon as the air clears a little, I'll be back on processing PRs here! Looking forward to checking out your improvements!

niku-live:
Hi Jesper,

Do not worry too much, of course Fall Release comes first. I just had some free time, so decided to update my PRs. We need those wrappers for our add-on but it can wait, there are still lots of converison bugs left besides .NET components, so we are not in an immediate need for them, we can wait for some time.

JesperSchulz:
And in it goes :-) We had to find a new home for the helper functions, as we don't quite know yet where we want to put the new helpers and how to structure them. Hope that is ok. On to the next one! Have a fantastic weekend once you make it that far!

PR #83, opened 2018-07-10
System.IO.Compression.GZipStream wrapper

niku-live:
Inspired by MemoryStream wrapper. There are cases when you need to compress / decompress GZip stream and having this kind of wrapper solves many problems.

### DIFF
--- a/BaseApp/COD50006.txt
+++ b/BaseApp/COD50006.txt
@@ -1,0 +1,126 @@
+OBJECT Codeunit 50006 DotNet_GzipStream
+{
+  OBJECT-PROPERTIES
+  {
+    Date=18-07-10;
+    Time=21:23:58;
+    Modified=Yes;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      DotNetGZipStream@17024400 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.IO.Compression.GZipStream";
+      TextErrorReadNotSupported@17024401 : TextConst 'ENU=Stream cannot be read from';
+      TextErrorWriteNotSupported@17024402 : TextConst 'ENU=Stream cannot be written to';
+
+    [External]
+    PROCEDURE InitGzipStreamFromTempBlob@17024400(VAR TempBlob@1000 : Record 99008535;CompressMode@17024401 : Boolean);
+    VAR
+      BaseInStream@17024400 : InStream;
+      BaseOutStream@17024402 : OutStream;
+    BEGIN
+      IF CompressMode THEN
+        BEGIN
+          CLEAR(TempBlob.Blob);
+          TempBlob.Blob.CREATEOUTSTREAM(BaseOutStream);
+          InitGzipStreamFromOutStream(BaseOutStream);
+        END
+      ELSE
+        BEGIN
+          TempBlob.Blob.CREATEINSTREAM(BaseInStream);
+          InitGzipStreamFromInStream(BaseInStream);
+        END;
+    END;
+
+    [External]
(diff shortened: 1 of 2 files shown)


BaseApp/COD50006.txt
```diff
@@ -0,0 +1,126 @@
...
+  }
+  CODE
+  {
+    VAR
+      DotNetGZipStream@17024400 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.IO.Compression.GZipStream";
+      TextErrorReadNotSupported@17024401 : TextConst 'ENU=Stream cannot be read from';
+      TextErrorWriteNotSupported@17024402 : TextConst 'ENU=Stream cannot be written to';
+
+    [External]
+    PROCEDURE InitGzipStreamFromTempBlob@17024400(VAR TempBlob@1000 : Record 99008535;CompressMode@17024401 : Boolean);
```
JesperSchulz: Assuming you have read my comments in #82, you know what a helper function is and where it should go. Please isolate helpers in separate codeunits.

BaseApp/COD50006.txt
```diff
@@ -0,0 +1,126 @@
...
+      DotNetGZipStream := DotNetGZipStream.GZipStream(BaseStream, DotNetCompressionMode.Decompress);
+    END;
+
+    [External]
+    PROCEDURE InitGzipStreamFromOutStream@17024402(VAR BaseStream@1000 : OutStream);
+    VAR
+      DotNetCompressionMode@17024400 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.IO.Compression.CompressionMode";
+    BEGIN
+      DotNetGZipStream := DotNetGZipStream.GZipStream(BaseStream, DotNetCompressionMode.Compress);
+    END;
```
JesperSchulz: Ideally these two functions (In and Out) are just one with the direction specified as a parameter. That way we follow the signature of the constructor, which gives us the greatest degree of flexibility and the smallest chance of having breaking changes down the road.

BaseApp/COD50006.txt
```diff
@@ -0,0 +1,126 @@
...
+    PROCEDURE WriteToStream@8(VAR OutStream@1000 : OutStream);
+    BEGIN
+      IF NOT CanRead THEN
+        ERROR(TextErrorReadNotSupported);
+      DotNetGZipStream.CopyTo(OutStream);
+    END;
+
+    [External]
+    PROCEDURE WriteToTempBlob@17024401(VAR TempBlob@1000 : Record 99008535);
+    VAR
```
JesperSchulz: Helper function.

JesperSchulz: Well, you get the picture by now. Won't leave any more "helper function" comments in your code.

niku-live:
Hi,

I moved helper functions to helper codeunit. Still I had to leave InitGzipStreamFromInStream and InitGzipStreamFromOutStream, I can't merge them into one function, because InStream and OutStream is not assignable one from another. I could make .NET Stream class as BaseStream param type, but then I would be exposing .NET type, which I want to avoid. Waiting for you comments.

niku-live:
Hi,

I made a few changes to my PR:
1. Fixed all IDs and renamed some procedures.
2. Removed InitGZipStreamFromInStream and InitGZipStreamFromOutStream, now procedure "GZipStream" is used, this procedure lets construct stream for both compression and decompression and represents wrapped .NET class more closely.
3. Added IO.Stream wrapper codeunit. By wrapping IO.Stream we can now pass InStream, OutStream, MemoryStream and other types of streams to the same constructor procedure "GZipStream". Without this wrapper we need to split constructor into 2 procedures (as it was before).
4. Update MemoryStream wrapper to include procedure to cast it to IO.Stream wrapper. One test iliustrates how it is used.

PR #84, opened 2018-07-10
SmtpMessage wrapper

niku-live:
Even so there is standard NAV functionality for sending Smtp Mail, in certain cases it is useful to have direct access to SmtpMessage DotNet class. This wrapper provides such possibility, so I think it is useful addition to C/AL open library

### DIFF
--- a/BaseApp/COD50008.txt
+++ b/BaseApp/COD50008.txt
@@ -1,0 +1,126 @@
+OBJECT Codeunit 50008 DotNet_SmtpMessage
+{
+  OBJECT-PROPERTIES
+  {
+    Date=;
+    Time=;
+    Modified=Yes;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      DotNetSmtpMessage@17024400 : DotNet "'Microsoft.Dynamics.Nav.SMTP, Version=11.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.Microsoft.Dynamics.Nav.SMTP.SmtpMessage";
+
+    [External]
+    PROCEDURE CreateMessage@17024401();
+    BEGIN
+      DotNetSmtpMessage := DotNetSmtpMessage.SmtpMessage;
+    END;
+
+    [External]
+    PROCEDURE SetFromAddress@17024402(FromAddress@17024400 : Text);
+    BEGIN
+      DotNetSmtpMessage.FromAddress := FromAddress;
+    END;
+
+    [External]
+    PROCEDURE SetFromName@17024403(FromName@17024400 : Text);
+    BEGIN
+      DotNetSmtpMessage.FromAddress := FromName;
+    END;
+
+    [External]
+    PROCEDURE SetToAddress@17024405(ToAddress@17024400 : Text);
+    BEGIN
+      DotNetSmtpMessage."To" := ToAddress;
+    END;
+
+    [External]
+    PROCEDURE SetSubject@17024410(Subject@17024400 : Text);
+    BEGIN
+      DotNetSmtpMessage.Subject := Subject;
+    END;
+
+    [External]
+    PROCEDURE SetAsHtmlFormatted@17024408(HtmlFormatted@17024400 : Boolean);
+    BEGIN
+      DotNetSmtpMessage.HtmlFormatted := HtmlFormatted;
+    END;
+
+    [External]
+    PROCEDURE SetTimeout@17024424(Timeout@17024400 : Integer);
+    BEGIN
+      DotNetSmtpMessage.Timeout := Timeout;
+      DotNetSmtpMessage.ConvertBase64ImagesToContentId
+    END;
+
+    [External]
+    PROCEDURE ClearBody@17024412();
+    BEGIN
+      DotNetSmtpMessage.Body := '';
+    END;
+
+    [External]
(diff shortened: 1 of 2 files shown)


JesperSchulz:
Thanks a lot for your contribution! ...or all of your contributions! I will look into them before end of this week! But at first glance, they look very clean. If I have any questions, I'll get back to you. Thanks again! Much appreciated!

JesperSchulz:
Your request was just discussed in a bigger forum where it was concluded that the functions in this request cannot be exposed for extension development; or rather the Send() function can't. This code would make it impossible to control the outflow of data out of Business Central; e.g. would it be impossible to prevent mails from being send from a Sandbox or test environment.
We are aware, that our mailing capabilities are sub-optimal and will invest in getting those right - also for extension development.
Would you like to move forward with this wrapper with an [Internal] Send function?

niku-live:
Hi, 

I understand the problem, yes I think Send function can be removed from the wrapper (there is no need to keep it as internal). Still for this wrapper to be useful, some kind of method to send message is still needed. It is just a thought but what about extending SMTP Mail Codeunit (ID 400) with function to update values from SmtpMail wrapper. Then Mail sending will be handled with default NAV Mail functionality, but mail mesage could be constructed using this wrapper. What do you think about such solution?

grobyns:
Hi,

we will take a broader look at how we send email from BC. We have several ways of sending mail in BC already and we need to clean that up.
Unfortunately that will take a while and we need to put this submission on hold until then.

br/Gert

niku-live:
Hi,

I understand situation, so I don't mind waiting. But maybe you could give us some insights of what we should expect from the future email functionality and what is the best way to refactor our own code to get ready for it.

grobyns:
Hi,

it's a little hard to comment on this since we haven't completed the investigation yet. I think we'll need to combine all the various implementations and places where we send email today to end up with a single point of setup and api for creating and sending email but I don't know what that will look like.

br/Gert

JesperSchulz:
We've decided to accept this wrapper afterall, as it does not make things worse than they are today. I'll process the code for check-in right away :-)

niku-live:
Hi Jesper,
It's good news as we can go further with our solution. 

Still looking further into the near future, I think we all hope for NAV/BC become better and better and for that let me share a few ideas how we could approach mail functionality (it's just my ideas but maybe it will be useful). So my view to future mail functionality would be something like this:
1. Actual sender code is decoupled from mail body/receivers construction code, we have different objects for mail construction and different for sending.
2. Mail construction could be implemented like that:
   2.1. For receivers list we have a buffer table "Mail Receiver Buffer" with fields Type (To, CC, BCC), "Receiver Address", "Receiver Name" and function to add new receiver or semicolon separated receiver list
  2.2. For attachments we have a buffer table "Mail Attachment Buffer" with fields "Attachment Blob", "Attachment Description" and maybe some other fields.
  2.3. Finally we have one or two buffer tables for main params (Mail Header and maybe Mail Body).
3. For sending code I think it is important to notice that there are different ways to send mail (Smtp, Outlook/Office 365, IMAP and maybe others). So maybe approach could be via "Discovery event" pattern:
  3.1. There is an event OnRegisterMailServerProvider which is used to register possible senders. By default NAV register Smtp sender. Every sender provider registers it's own setup page, so you can have different configurations for different providers. For cloud deployments we can limit and forbid registering new providers, for on-premise deployment it could be allowed.
  3.2. Sending is done through sender wrapping codeunit Mail Sender. Codeunit accepts Mail Body, Mail header, Mail Receiver and Mail Attachment tables as parameters and depending on system configuration sends mail using right sender provider. It could also be allowed to give a hint to the wrapper which provider to use (for example if you must use Outlook provider)
  3.3. Actual sender codeunit (e.g. Smtp Sender, Outlook sender and etc.) has no public members just OnRun trigger that accepts parameters table (for example it could be a table called Mail Parameter Composite, that can hold parameters from all other param tables - Mail Attachment, Mail Body, Mail header, Mail Receiver and etc. or it could be different approach). Mail Sender wrapper basically just calls Codeunit.Run(MailProviderSetup."Codeunit ID", MailParameters) depending on configuration.

Well this is just a rough idea, It should be improved a lot, but as I understand mail sending functionality is being redesigned so just my two cents for consideration.

BaseApp/COD3031.txt
```diff
@@ -0,0 +1,125 @@
...
+    PROCEDURE SetAsHtmlFormatted@6(HtmlFormatted@1000 : Boolean);
+    BEGIN
+      DotNetSmtpMessage.HtmlFormatted := HtmlFormatted;
+    END;
+
+    [External]
+    PROCEDURE SetTimeout@7(Timeout@1000 : Integer);
+    BEGIN
+      DotNetSmtpMessage.Timeout := Timeout;
+      DotNetSmtpMessage.ConvertBase64ImagesToContentId;
```
JesperSchulz: Is this line supposed to be there?

JesperSchulz:
Thanks for all your valuable input! We will consider that, when looking into BCs mail sending functionality!
But first things first: Let's get this wrapper into our depot! I found one line of code, which I'm not sure if it's there on purpose. Could you confirm (I left a comment)? Then I'll wrap this up and submit a build job :-)

BaseApp/COD3031.txt
niku-live: Oh. Yes it shouldn't be here :) Well you what happens with copy-paste.

JesperSchulz: :-D I'll remove it and move forward with this request. Thanks!

PR #85, opened 2018-07-12
IEnumerable, IEnumerator, System.Object and System.Type wrapper

niku-live:
I agree that this commit might be close to the limits of what should be allowed in C/AL Open Library, but I think this addition is needed and very useful, so let me explain everything. Basically all included objects is for IEnumerator support.
Almost all collections, lots of default methods and and etc. supports enumerators in .NET. So wrapper for IEnumerator is really needed to access this functionality. But the problem comes here - to support it in a generic way, we also need wrappers for System.Object and System.Type. System.Object wrapper allows to wrap the result of IEnumerator.Current and System.Type allows us to check the type name and whether type is assignable from. I tried making only safe methods of System.Type visible, so from my point of view it should be safe wrapper. Also I extended Array wrapper - it now can be converter to IEnumerator wrapper and we can also create it (using Type wrapper) and later modify.

### DIFF
--- a/BaseApp/COD3000.TXT
+++ b/BaseApp/COD3000.TXT
@@ -4,6 +4,7 @@ OBJECT Codeunit 3000 DotNet_Array
   {
     Date=;
     Time=;
+    Modified=Yes;
     Version List=;
   }
   PROPERTIES
@@ -18,15 +19,48 @@ OBJECT Codeunit 3000 DotNet_Array
       DotNetArray@1000 : DotNet "'mscorlib'.System.Array";
 
     [External]
+    PROCEDURE Create@17024402(VAR ElementType@17024401 : Codeunit 50011;Length@17024400 : Integer) : Integer;
+    VAR
+      DotNetType@17024402 : DotNet "'mscorlib'.System.Type";
+    BEGIN
+      ElementType.GetType(DotNetType);
+      DotNetArray := DotNetArray.CreateInstance(DotNetType, Length);
+    END;
+
+    [External]
     PROCEDURE Length@4() : Integer;
     BEGIN
       EXIT(DotNetArray.Length)
     END;
 
     [External]
+    PROCEDURE SetValueAsObject@17024407(Index@1000 : Integer;VAR NewValue@17024400 : Codeunit 50010);
+    VAR
+      DotNetObject@17024401 : DotNet "'mscorlib'.System.Object";
+    BEGIN
+      NewValue.GetObject(DotNetObject);
+      DotNetArray.SetValue(DotNetObject, Index);
+    END;
+
+    [External]
+    PROCEDURE SetValueAsText@17024409(Index@1000 : Integer;NewValue@17024400 : Text);
+    VAR
+      DotNetString@17024401 : DotNet "'mscorlib'.System.String";
+    BEGIN
+      DotNetString := NewValue;
+      DotNetArray.SetValue(DotNetString, Index);
+    END;
+
+    [External]
     PROCEDURE GetValueAsText@5(Index@1000 : Integer) : Text;
     BEGIN
       EXIT(DotNetArray.GetValue(Index))
+    END;
+
+    [External]
+    PROCEDURE GetValueAsObject@17024405(Index@1000 : Integer;VAR CurrentValue@17024400 : Codeunit 50010);
+    BEGIN
+      CurrentValue.SetObject(DotNetArray.GetValue(Index));
     END;
 
     PROCEDURE GetArray@2(VAR DotNetArray2@1000 : DotNet "'mscorlib'.System.Array");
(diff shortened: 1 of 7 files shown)


JesperSchulz:
Fantastic! Thanks for the detailed explanation - makes reviewing the code so much simpler. I'll take this PR to the triage meeting today, but I'm optimistic that we'll accept this one :-) Thank you so much for contributing. I'll try to turn around your requests swiftly!

JesperSchulz:
We have decided to not accept this particular submission. We think the solution is too generic and would send this initiative down the wrong path. We would prefer to put specific methods for enumerations directly in the wrappers as the need occurs. I hope you understand and will keep submitting the wrappers you need to create great extensions with us!

Have a great weekend!

niku-live:
Hi, I fully understand and even I myself wrote that this is close to the limits what should be allowed. So it is okay, to reject this solution :). Still I would like to discuss a little bit to understand better vision of were C/AL open library should go. So it will be long most, but please hear me out. In defense of generic enumerators - my next wrapper pack that is coming wraps various classes from DocumentFormat.OpenXml namespace. If you look at Document.OpenXml you can notice that classes has lots of collections (ChildElements, DocumentParts, StyleDefinitions, Relationships and so on). Most of the time what you will need when working with such classes - items enumeration. True, I can create iterator methods for every wrapper, but then it will be impossible to create helpful helper codeunits dedicated to enumeration handling (oh I so wish NAV had codeunit XX implements YY interface construct).  I have enough C# background to say that IEnumerable interface lets you forget a lot and create generic solutions with clear and simple code. Absence of generic solutions is what NAV lacks a lot. Of course everything comes with a price, basically its generalization vs simplification. At TechDays Microsoft even proposed Variant Facade design pattern that uses kind similar level of generalization (Variant == Obejct Wrapper). I understand that primary goal of C/AL open library - help accelerating migration to extensions. But how we achieve that is another question. I am just one developer so of course my opinion is just one point of view but lets say we try to avoid generic solutions, now we take Array Wrapper as an example. For now is always returns element value as string. Sometime in the future anothe developer need an array of integers (because some class method needs int array), so another wrapper is created "Int Array Wrapper", then someone creates "Double Array Wrapper" and so on and then we have lots of wrappers that does the same thing but with different types. One generic solution with Object wrapper and some type helpers could solve this more beautifully. Yes, generic solutions might be harder to understand, but since C/AL Open Library deals with .NET that is kinda advanced topic for NAV developers, so I expect that audience of this repository can handle generic solutions easily. So my question is - how far contributors can go with generic solutions and what we shuld avoid.

JesperSchulz:
Sorry about the late reply. I’m currently on a summer break far away from computers :-) I will forward your question to one of my colleagues. And I’ll of course be back on the case when I’m back from vacation! Talk to you soon!

grobyns:
Hi,

the purpose of the C/AL Open Library is to unblock partners who are moving their solutions to the cloud by providing a temporary workaround. Ideally the AL language would provide the capabilities required but it takes time to develop those so in the meantime, partners can suggest .NET wrappers here to quickly unblock them.
The C/AL Open Library is only supposed to provide the methods actually in use by the partner which is why the methods are specific and we do not implement all the methods in a class. We also want it to be easy to move away from these wrappers as the AL language gains more capabilities.

Regarding generic solutions: We have to be careful with what we allow in these wrappers. The more generic a solution, the harder it becomes to foresee how it will be used and what its impact will be.   
Trying to replicate .NET code 1:1 in AL through wrappers is not the right approach. For more generic and heavier .NET usages you should consider implementing an Azure Function and calling it from AL.  C/AL Open Library should be thought of as a mitigation to an acute problem and not a place to build-up generic, long-term solutions. Another way to put it is – if you have legacy code that will be significantly difficult to migrate to Business Central because of the lack of a few specific .NET scenarios, C/AL Open Library is the right place to solve it. If you are writing new AL code or have broad usage of .NET you should consider moving to native AL types or using Azure Functions (or other services).

br/Gert

niku-live:
Hi,

It is just a thought, but how about creating "AL Open Library" with different purpose than "C/AL Open Library" - to create community supported AL Library extension with unblocked .NET functions and various helper functions. We already have .NET support for on-premise AL developments, but this feature will never be supported on cloud, so if we could propose various .NET extensions to this new github repository, we could have a special library extension with longer life span than just a "temporary workaround". It could be kinda like "NuGet for NAV". I understand that AL language will be extended in the future, but I doubt that so many new data types will be added as standard .NET framework provides. Azure functions is a solution in this case but this solution fails when we need to deal with bigger data streams or perform many different .NET method calls (for example creating word documents using DocumentFormat.OpenXml library). Yes we cannot use AL extensions with .NET in the cloud, but Microsoft can, so Microsoft can create open source AL extension with .NET methods, community then could propose pull requests to it, in the end we will have powerful but managed useful functions library.

PR #88, opened 2018-08-06
Dot net string ends starts with

gunnargestsson:
Add functions: StartsWith and EndsWith

### DIFF
--- a/BaseApp/COD3007.TXT
+++ b/BaseApp/COD3007.TXT
@@ -18,21 +18,18 @@ OBJECT Codeunit 3007 DotNet_String
       DotNetString@1000 : DotNet "'mscorlib'.System.String";
 
     [External]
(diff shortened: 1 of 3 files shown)


BaseApp/COD3007.TXT
```diff
@@ -2,8 +2,9 @@ OBJECT Codeunit 3007 DotNet_String
 {
   OBJECT-PROPERTIES
   {
-    Date=;
-    Time=;
+    Date=23.07.18;
```
JesperSchulz: Try to avoid unnecessary changes which distract from the code you'd like us to review, such as adding date/time or the LineStart below; both of which shouldn't be submitted to the repo ;-)

gunnargestsson:
Cleaned

PR #89, opened 2018-08-06
Dot net uri

gunnargestsson:
Added functionality


BaseApp/COD3024.TXT
```diff
@@ -2,9 +2,10 @@ OBJECT Codeunit 3024 DotNet_Uri
 {
   OBJECT-PROPERTIES
   {
-    Date=;
-    Time=;
-    Version List=;
+    Date=16.07.18;
```
JesperSchulz: Same comment as in the previous PR. Please remove all code changes we don't need, such as Date/Time and the LineStart. Ideally, you should also not update the .gitignore file of this repo; keep those changes local.

gunnargestsson:
Cleaned

PR #90, opened 2018-08-06
Dot net collection

gunnargestsson:


JesperSchulz:
This seems to be overlapping with the native implementation of lists in the al language? We should not have competing implementations. If you still need collections, they should be [Internal]. Do you agree?

JesperSchulz:

PR #92, opened 2018-08-14
Dot net web client

gunnargestsson:
Added a security check for the Url.


grobyns:
Hi Gunnar,

Why do we need this? Isn't the HttpClient functionality in al capable of doing the same? If it is not, then that feedback should go to the modern dev team. 
Do you need these methods in SaaS or is it ok to keep them as Internal for now?

br/Gert

BaseApp/COD3041.TXT
```diff
@@ -0,0 +1,125 @@
...
+    LOCAL PROCEDURE DownloadString@7(Url@1000 : Text) : Text;
+    BEGIN
+      EXIT(DotNetWebClient.DownloadString(Url));
+    END;
+
+    LOCAL PROCEDURE VerifyUrl@6(Url@1000 : Text);
+    VAR
+      DotNet_Uri@1001 : Codeunit 3024;
+    BEGIN
+      DotNet_Uri.Init(Url);
```
grobyns: This method breaks the dotnet wrapper rules but as it is a security thing that's ok.

We have a bunch of similar functions in Codeunit 1299. In addition to http/https it verifies whether the url is well formed and valid. You could copy that functionality here.

BaseApp/COD3041.TXT
```diff
@@ -0,0 +1,125 @@
...
+    BEGIN
+      DotNetWebClient2 := DotNetWebClient;
+    END;
+
+    PROCEDURE SetWebClient@2(DotNetWebClient2@1000 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Net.WebClient");
+    BEGIN
+      DotNetWebClient := DotNetWebClient2;
+    END;
+
+    EVENT DotNetWebClient@1000::WriteStreamClosed@105(sender@1001 : Variant;e@1000 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Net.WriteStreamClosedEventArgs");
```
grobyns: Do we need these events or is ok to remove them from the source?

BaseApp/COD3041.TXT
gunnargestsson: You still want me to use a codeunit that is not a wrapper and containes dotnet object?

We can do

WebRequestHelper.IsValidUri(Url);
IF NOT (WebRequestHelper.IsSecureHttpUrl(Url) OR WebRequestHelper.IsHttpUrl(Url)) THEN
  ERROR(UrlSecurityErr);

grobyns: No, the wrapper cannot take a dependency on a non-wrapper object. I was merely suggesting you copy paste the code so that the check is the same in both places.

gunnargestsson:
Files updated based on suggestions

grobyns:
Shouldn't that functionality be built into the modern al language instead?

br/Gert

StanislawStempin:
AL already contains a set of HTTP* types based on the newer HttpClient .NET class stack. You should be able to achieve any network scenarios using these new types. If not, we'd love to hear about the gaps.
We are not planning to introduce WebClient into AL.

gunnargestsson:
My Soap Proxy Client is using dotnet to download WSDL and build an Assembly in memory.  Using reflection to then execute methods.  http://www.dynamics.is/?p=3032
When I have sent all the wrappers I will be able to do this and use the same code for both C/AL and AL

grobyns:
Hi Gunnar,

I don't see us allowing what you propose in SaaS and if we don't need it in SaaS we wouldn't put the wrappers in the product.

br/Gert

gunnargestsson:
Thank you Gert.
Then I don't see the need for my company to send any more wrappers at this time.

gunnargestsson:
Here are some of the endpoints that I need to work with.  

This is banking.

https://ws.b2b.is/services/

This is IRS

https://vefurp.rsk.is/ws/Stadgreidsla/StadgreidslaService.svc
https://thjonusta-s.rsk.is/virdisaukaskattur/skil/VSKService?WSDL

Will I be able to use these endpoints from AL Http* variables?

StanislawStempin:
I'm not familiar with these services. In general you should be able to call all services from the new AL types.
If something is not possible please let us know.

gunnargestsson:
In short, I am forced to move to Azure Functions.

JesperSchulz:

PR #94, opened 2018-08-16
StreamReader DotNet wrapper

niku-live:
It seems there is a StreamWriter wrapper but no StreamReader wrapper, so I want to change this. For our custom add-on StreamReader is really important because we need to import text files created using many different encodings (our government institutions still use applications that do not support Unicode) and using NAV TextEncoding property / data type is not enough (because there is no way to use Windows-1257, Windows-1252, Windows-1250, OEM-775, ISO 8859-4, ISO 8859-13 and ISO 8859-10 encodings in one application configuration). I hope that in future Microsoft will extend existing encoding functionality and make it more useful but for now we really need StreamReader.

This commit contains StreamReader wrapper codeunit and a few extensions to existing wrappers:
1) StreamWriter wrapper extended with Write function as sometimes you need to write text without new line character
2) Encoding wrapper extended with a function to get what is current codepage and also with methods to construct is from static properties (Utf8, Unicode and etc.) because we use it a lot when creating Unicode based encoding
3) String wrapper extended with functions ToText (strange that no one needed it, how do you get what is holds?) and IsEmpty (basically to check if it is null, needed when using ReadLine from StreamReader)


JesperSchulz:
Cool :-) I'll look into this one right away.

JesperSchulz:
Merge pull request #94 from niku-live/StreamReaderWrapper

I had to address a few internal code review comments, before accepting this PR. But the checked in solution is close to your suggestion, so I think it's ok. If not, please open a new PR.

PR #95, opened 2018-08-22
Array, String and Encoding wrappers extended with new methods

niku-live:
Array wrapper extended with methods to create and manipulate Byte, Char, Int32 and String arrays (all 4 types of array is frequently used, especially when working with streams and text).

String wrapper extended with new methods to manipulate text. I know, that AL has its own Text datatype with similar methods, but these few methods I added are frequently used and it is very strange to always convert to/from DotNetString wrapper just to call native methods.

Encoding wrapper extended with methods to get string directly from byte array, this is very useful when implementing binary data file formats import functionality.

Test codeunits iliustrate how to use all of those new methods.


GiorgosPa:
Thanks for the suggestion! We’re investigating whether we can use it and hope to provide an answer soon.

Test/COD50300.TXT
```diff
@@ -0,0 +1,153 @@
+OBJECT Codeunit 50300 Test_DotNet_Encoding
```
GiorgosPa: The tests should be moved in Codeunit 146026

Test/COD50302.TXT
```diff
@@ -0,0 +1,161 @@
+OBJECT Codeunit 50302 Test_DotNet_String
```
GiorgosPa: Please re-id to 146007

Test/COD50301.TXT
```diff
@@ -0,0 +1,143 @@
+OBJECT Codeunit 50301 Test_DotNet_Array
```
GiorgosPa: Please re-id to 146000

BaseApp/COD3000.TXT
```diff
@@ -17,18 +17,100 @@ OBJECT Codeunit 3000 DotNet_Array
...
+    BEGIN
+      EXIT(DotNetArray.GetValue(Index));
+    END;
+
+    [External]
+    PROCEDURE GetValueAsInteger@106(Index@1000 : Integer) : Integer;
+    BEGIN
+      EXIT(DotNetArray.GetValue(Index));
+    END;
+
```
GiorgosPa: Add A GetValueAsByte for completenes

Test/COD50300.TXT
niku-live: Fixed with latest commits

BaseApp/COD3000.TXT
niku-live: Fixed with latest commits

Test/COD50301.TXT
niku-live: Fixed with latest commits

Test/COD50302.TXT
niku-live: Fixed with latest commits

niku-live:
Hi,

With my latest commits I moved Test codeunits to correct IDs also implemented GetValueAsByte for completeness. About case sensitive problems - I didn't even noticed that in my repository files were with wrong case. I tried fixing this by removing files and adding them again. At least on my local repository now everything look clear and ok, but write me if something is wrong.

PR #96, opened 2018-08-23
BinaryStreamWriter, BinaryStreamReader, IO.Stream and IO.Buffer wrappers

niku-live:
This PR includes BinaryStreamWriter, BinaryStreamReader, IO.Stream and IO.Buffer wrappers. Having those wrappers enpowers NAV developers to implement custom binary file formats in NAV without depending on Azure functions (so avoiding performance issues). 

For example, in our custom add-on we support import/export from DBF files. Originally we used third-party library FastDBF, but we cannot use it in cloud deployments, so we reimplemented whole library in C/AL (source code here: https://github.com/niku-live/FastDBF-NAV), however we still need BinaryStreamReader/Writer to work directly with bytes (as NAV do not have Byte datatype)


Test/COD50303.TXT
```diff
@@ -0,0 +1,424 @@
...
+      DotNet_BinaryReader.Dispose;
+      DotNet_BinaryWriter.Dispose;
+    END;
+
+    [Test]
+    PROCEDURE TestWriteBool@2();
+    BEGIN
+      //Based on https://github.com/dotnet/corefx/ BinaryWriter.WriteTests.cs BinaryWriter_WriteBoolTest
+      DotNet_MemoryStream.InitMemoryStream;
+      DotNet_MemoryStream.CastToStream(DotNet_Stream);
```
mazhelez: The code is not compiling.
`CastToStream` is not a method/procedure in Codeunit 3009, the type of `DotNet_MemoryStream` :)

Please, provide the missing method.

mazhelez: Just saw that the required changes are in another PR.

I will look into it now :)

niku-live: Hello,

Yes, it seems I forgot, that I used DotNet_MemoryStream modifications from another PR. Originally I hoped, that first PR will be approved before processing this one, well I was wrong :). Since CastToStream is used only for tests, I think I can modify test code to avoid using this function.

mazhelez: No, it's not necessary 🙂

I'll have a look at your other PR and then proceed with this one.

In such cases, if possible, note such dependencies so that the PRs are processed in the right order.

PR #97, opened 2018-08-28
Wrappes needed to encrypt and decrypt text with a key

gunnargestsson:
Plus two new functions to DotNet_Encoding

### DIFF
--- a/BaseApp/COD53080.TXT
+++ b/BaseApp/COD53080.TXT
@@ -1,0 +1,39 @@
+OBJECT Codeunit 53080 DotNet_MD5 Crypto Service
+{
+  OBJECT-PROPERTIES
+  {
+    Date=;
+    Time=;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+    VAR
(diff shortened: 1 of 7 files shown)


BaseApp/COD53080.TXT
```diff
@@ -0,0 +1,39 @@
+OBJECT Codeunit 53080 DotNet_MD5 Crypto Service
```
JesperSchulz: I think we've retired support for MD5 due to vulnarabilities. Is this needed?

gunnargestsson: We need to be able to encrypt and decrypt using a key and do the same thing in C#.  MD5 is not required, just a method that works and we can trust.

As we send passwords inside a Json to Azure Function they are encrypted with a key and the Azure Function has the key and the function to decrypt the password before passing to the actual Soap service.

JesperSchulz:
Hi Gunnar!

I need to clean out our C/AL Open Library and this one remains unresolved. Could I ask you to check, if the https://github.com/microsoft/ALAppExtensions/tree/master/Modules/Cryptography%20Management module will do the trick? If it doesn't, would you mind telling us what more you need in that module to support your scenario? I will close this PR, as we won't be adding functionality to the wrappers, now that we have a cryptography module. Thanks for helping out on validating, that we support everything that is needed :-)

PR #98, opened 2018-08-30
Regex wrapper

jakubhojnacki:
A set of codeunits wrapping .NET Regex classes in C/AL.
An idea is to treat codeunits as equivalents of .NET classes 1:1 to make it as close to .NET as possible.
Further comments exist in objects "Documentation" trigger.
Full set of test codeunits has been provided.

### DIFF
--- a/BaseApp/COD66601.TXT
+++ b/BaseApp/COD66601.TXT
@@ -1,0 +1,182 @@
+OBJECT Codeunit 66601 RegexOptions
+{
+  OBJECT-PROPERTIES
+  {
+    Date=;
+    Time=;
+    Version List=;
+  }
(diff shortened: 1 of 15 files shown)


BaseApp/COD66601.TXT
```diff
@@ -0,0 +1,182 @@
+OBJECT Codeunit 66601 RegexOptions
```
JesperSchulz: Please follow our coding guidelines with regards to naming etc. Those guidelines were very well hidden, for which I apologize. I've now moved them to the frontpage.

JesperSchulz:
PR was replaced by version 2.0 in #99. This PR is closed.

PR #99, opened 2018-09-04
Regex 2.0 wrapper

jakubhojnacki:
Verified and designed according to the guidelines (I hope so).
Two existing wrappers have been extended (DotNet_Array and DotNet_Regex) and a number of others added.
DotNet_RegexOptions have some deviations from the guidelines because it implements piped / bitwise manipulation of option values.

### DIFF
--- a/BaseApp/COD3041.TXT
+++ b/BaseApp/COD3041.TXT
@@ -1,0 +1,156 @@
+OBJECT Codeunit 3041 DotNet_RegexOptions
+{
+  OBJECT-PROPERTIES
+  {
+    Date=;
+    Time=;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      DotNetRegexOptions@1000000001 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Text.RegularExpressions.RegexOptions";
+      DotNetRegexOptionsValues@1000000000 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Collections.Generic.Dictionary`2";
+
+    [External]
+    PROCEDURE AddCompiled@1000000009();
+    BEGIN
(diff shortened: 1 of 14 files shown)


BaseApp/COD3041.TXT
```diff
@@ -0,0 +1,156 @@
...
+      EXIT(String);
+    END;
+
+    PROCEDURE GetRegexOptions@1000000002(VAR DotNetRegexOptions2@1000000000 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Text.RegularExpressions.RegexOptions");
+    VAR
+      String@1000000002 : Text;
+    BEGIN
+      DotNetRegexOptions := DotNetRegexOptions.None;
+      String := ToString();
+      IF (String <> '') THEN BEGIN
```
mazhelez: Hi :)

Can you please explain why the Get and Set methods diverged from what they should be?

BaseApp/COD3041.TXT
```diff
@@ -0,0 +1,156 @@
+OBJECT Codeunit 3041 DotNet_RegexOptions
```
mazhelez: This isn't a .Net wrapper, according to the guidelines.

Check [cod3037 ](https://github.com/Microsoft/cal-open-library/blob/master/BaseApp/COD3037.TXT) for reference on how to implement a .Net wrapper for enums.

jakubhojnacki: Hi Maria

The codeunit 3037 doesn't implement enums as they are implemented in .NET.
For RegexOptions - in .NET - you can do this:
  RegexOptions.Compiled | RegexOptions.IgnoreCase
Because the way Regex is designed, the feature above is very frequently used (I, myslef, use these options with pipes).
That's why I have added a number of methods allowing you to COMBINE options, not only use one of them.
Please, tell me how to implement it in NAV in any different way I did - I will happily change the code.
Best regards
Jacob

jakubhojnacki: Hi Maria
Thank you for your suggestion. I'm amending it as we speak; I'll publish a new pull request as soon as I have it ready.

JesperSchulz:
Closed as per request. Replaced by #107.

PR #104, opened 2018-09-18
ExcelBuffer with capability work with more than one sheet when reading file

niku-live:
This PR aims to extend ExcelBuffer with capability to read multiple sheets Excel files without continously reopening file. Originally ReadSheet closes file after reading one sheet, so I extended Excel buffer with the following functions:
ReadSheetContinous - basically does the same as ReadSheet but has options to select if files should be closed on completion and to select sheet name.
SetActiveReaderSheet - lets change currently used sheet for reader.
GetSheetsNameListFromStream - returns list of sheet names in file stream.
SaveToStream - somehow there are fuctions to open Excel file from stream, but no way to save it back to stream. Sometime we do not want to return file directly to user (DOWNLOAD function), but for example to save it to Blob. This is when this function becomes useful.


GiorgosPa:
Hi there! Glad to see that you’d like to collaborate. We have received your request and we’re looking at it now, we’ll get back to you shortly with a go/no-go decision.

PR #107, opened 2018-10-10
Regex 2.01 wrapper

jakubhojnacki:
Extension of existing Regex with RegexOptions redesigned.
Plus matching test scripts.
Replacement for "Regex 2.0 wrapper" pull request which can be closed.

### DIFF
--- a/BaseApp/COD3051.TXT
+++ b/BaseApp/COD3051.TXT
@@ -1,0 +1,209 @@
+OBJECT Codeunit 3051 DotNet_RegexOptions
+{
+  OBJECT-PROPERTIES
+  {
+    Date=;
+    Time=;
+    Version List=;
+  }
+  PROPERTIES
+  {
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      DotNetRegexOptions@1000000001 : DotNet "'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Text.RegularExpressions.RegexOptions";
+
+    [External]
+    PROCEDURE Compiled@1000000009() : Integer;
+    BEGIN
+      DotNetRegexOptions := DotNetRegexOptions.Compiled;
+      EXIT(DotNetRegexOptions);
+    END;
+
+    [External]
+    PROCEDURE CultureInvariant@1000000017() : Integer;
+    BEGIN
+      DotNetRegexOptions := DotNetRegexOptions.CultureInvariant;
+      EXIT(DotNetRegexOptions);
+    END;
+
+    [External]
+    PROCEDURE ECMAScript@1000000018() : Integer;
+    BEGIN
+      DotNetRegexOptions := DotNetRegexOptions.ECMAScript;
+      EXIT(DotNetRegexOptions);
+    END;
+
+    [External]
+    PROCEDURE ExplicitCapture@1000000019() : Integer;
+    BEGIN
+      DotNetRegexOptions := DotNetRegexOptions.ExplicitCapture;
+      EXIT(DotNetRegexOptions);
+    END;
+
+    [External]
+    PROCEDURE IgnoreCase@1000000020() : Integer;
+    BEGIN
+      DotNetRegexOptions := DotNetRegexOptions.IgnoreCase;
+      EXIT(DotNetRegexOptions);
+    END;
+
+    [External]
+    PROCEDURE IgnorePatternWhitespace@1000000021() : Integer;
+    BEGIN
+      DotNetRegexOptions := DotNetRegexOptions.IgnorePatternWhitespace;
+      EXIT(DotNetRegexOptions);
+    END;
+
+    [External]
+    PROCEDURE Multiline@1000000022() : Integer;
+    BEGIN
+      DotNetRegexOptions := DotNetRegexOptions.Multiline;
+      EXIT(DotNetRegexOptions);
+    END;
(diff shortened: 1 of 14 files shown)


JesperSchulz:
Thanks for your submission! We'll soon look into it! Stay tuned :-)

JesperSchulz:
We haven't forgot this one, but we're currently looking into a larger effort to expose a few, much used, .NET libraries. RegEx is part of that effort and there hence is quite some overlap. A little more patience, while we make sure we cover your needs and avoid duplicate functionality.

BaseApp/COD3051.TXT
```diff
@@ -0,0 +1,209 @@
...
+
+    [External]
+    PROCEDURE Singleline@1000000025() : Integer;
+    BEGIN
+      DotNetRegexOptions := DotNetRegexOptions.Singleline;
+      EXIT(DotNetRegexOptions);
+    END;
+
+    [External]
+    PROCEDURE BitwiseOr@100000016(Option1@100000000 : Integer;Option2@100000001 : Integer) : Integer;
```
mazhelez: Unfortunately, we cannot accept this kind of functions in a .Net wrapper.

You can define it in your own code, or use + in this case.
`RegexOptions` values are power of two so summing will have the same effect as bit-wise or.

Test/COD145900.TXT
```diff
@@ -0,0 +1,27 @@
+OBJECT Codeunit 145900 Test_DotNet_Regex_Runner
```
mazhelez: No need for that codunit 😄

Test/COD145901.TXT
```diff
@@ -0,0 +1,677 @@
...
+  {
+    Subtype=Test;
+    OnRun=BEGIN
+          END;
+
+  }
+  CODE
+  {
+    VAR
+      TestToolkit@1000000000 : Codeunit 145904;
```
mazhelez: Please, use the Assert (codeunit 130000). No need to reinvent it.

Test/COD145903.TXT
```diff
@@ -0,0 +1,87 @@
...
+    PROCEDURE TestGetValueAsText@1000000000();
+    VAR
+      Pattern@1000000000 : Text;
+      ExpectedNames@1000000001 : DotNet "'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.System.Collections.ArrayList";
+    BEGIN
+      Pattern := '(?<Family>[A-Z]\d{2}[A-Z])-(?<Series>\d{3})-(?<Variant>[A-Z])';
+      TestToolkit.SetArray4(ExpectedNames, '0', 'Family', 'Series', 'Variant');
+      TestGetValueAsTextOnce('Part No.: A23B-234-C, Replacement: C99Z-902-M', Pattern, 4, ExpectedNames);
+      TestGetValueAsTextOnce('Part No.: 123B-234-C, Replacement: C99Z-902-M', Pattern, 4, ExpectedNames);
+    END;
```
mazhelez: New line should follow here.

Test/COD145904.TXT
```diff
@@ -0,0 +1,186 @@
...
+    PROCEDURE VerifyTextResult@1000000006(Arguments@1000000003 : Text;Result@1000000001 : Text;ExpectedResult@1000000000 : Text);
+    BEGIN
+      VerifyResult(Arguments, Result = ExpectedResult, Result, ExpectedResult);
+    END;
+    PROCEDURE VerifyResult@1000000003(Arguments@1000000003 : Text;Verified@1000000001 : Boolean;Result@1000000000 : Variant;ExpectedResult@1000000004 : Variant);
+    BEGIN
+      IF ( NOT (Verified)) THEN
+        ERROR(TxtError, Arguments, Result, ExpectedResult);
+    END;
+    PROCEDURE ToString1@1000000007(Value1@1000000000 : Variant) : Text;
```
mazhelez: This function only adds boilerplate.

Just use `STRSUBSTNO` with whatever number of arguments you want 😃

Test/COD145904.TXT
```diff
@@ -0,0 +1,186 @@
+OBJECT Codeunit 145904 Test_DotNet_Regex_Toolkit
```
mazhelez: This helper codeunit should not exist.

Most of the functions inside can be replaces with functions from the Assert codeunit, the others should be distributed to other test codeunits in this PR?

mazhelez:
Due to lack of activity, I'll now close the PR.

However, similar changes can be found in #118

jakubhojnacki:
Good afternoon Maria
I'm sorry I was quiet about the regex wrapper, but activity behind it took me off guard - I got a message from you (I mean - "you" as "Microsoft") this project would not be continued and we would get AL built-in regex support instead.
I'm glad somebody else pushed more complete set of regex libraries in, but I can't see any test libraries created for it.
Do you want me to do anything about it? Fix and complete my former set of tests working with new codeunits? Or am I missing something and the tests are there? :-)
Best regards
Jacob

JesperSchulz:
As Maria has gone on a well deserved vacation, I'll try to reply... You're right that we were looking into enhancing the language, but since the timeline is unclear, we've decided to move forward with a wrapper pack as temporary fix. We hope it also covers your needs.
Don't worry about the tests. As these are pure wrappers, there's no need for elaborate testing. We did not introduce any new logic here. So there's no need for you to do anything more, unless you're missing some functionality you'd like us to add :-)

PR #124, opened 2019-06-27
Update Base Objects.

JesperSchulz:


kwettermann:
Hi @JesperSchulz, were the changes from #119 (COD3010, COD3012) removed by mistake?

JesperSchulz:
Well... Kind of! We're currently juggling two versions: version 14 which is C/AL and the upcoming version 15, which will be in al. Today I updated the Base Objects with the ones from version 14, to which I didn’t backport the changes you are mentioning. Hence they appear gone again. They are, however, still part of version 15. This is all a little confusing and we need figure out what to do here. Sorry about that. Your changes are still in the upcoming release.

JesperSchulz:
@kwettermann: Well... Kind of! We're currently juggling two versions: version 14 which is C/AL (like this repo) and the upcoming version 15, which will be in al. Today I updated the Base Objects with the ones from version 14, to which I didn’t backport the changes you are mentioning - that was probably a mistake. Hence they appear gone again. They are, however, still part of version 15. But I cannot update with version 15 objects, as those are in al. I would have to redo this repo. This is all a little confusing and we need figure out what to do here. Sorry about that. Your changes are still in the upcoming release.

